PR #2, opened 2022-03-23
Circle spawner

josh65536:
Added a `PositionCircleModifier` so that circle-shaped emitters can be made.

### DIFF
--- a/src/modifiers.rs
+++ b/src/modifiers.rs
@@ -79,10 +79,7 @@ impl InitModifier for PositionCircleModifier {
             ShapeDimension::Volume => {
                 // Radius uniformly distributed in [0:1], then square-rooted
                 // to account for the increased perimeter covered by increased radii.
-                format!(
-                    "let r = sqrt(rand()) * {};",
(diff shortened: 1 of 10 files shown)


src/modifiers.rs
```diff
@@ -1,156 +1,236 @@
...
-    /// The constant acceleration to apply to all particles in the effect each frame.
-    pub accel: Vec3,
-}
-
-impl UpdateModifier for AccelModifier {
-    fn apply(&self, layout: &mut UpdateLayout) {
-        layout.accel = self.accel;
-    }
-}
+use bevy::prelude::*;
```
djeedai, line 1: What happened here? Bad case of CR/LF change?

djeedai (approved):

djeedai (commented):
Maybe add that new disk spawner to the feature list in `README.md`?

src/modifiers.rs
josh65536: Yeah, the newlines used to be CRLF and now they're just LF

PR #4, opened 2022-03-25
Spawn on command

josh65536:
Allows resetting, activation, and deactivation of spawners.

### DIFF
--- a/examples/circle.rs
+++ b/examples/circle.rs
@@ -84,17 +84,16 @@ fn setup(
     });
 
     // The sphere
-    commands
-        .spawn_bundle(PbrBundle {
-            mesh: meshes.add(Mesh::from(shape::UVSphere {
-                radius: 1.0,
-                sectors: 32,
-                stacks: 16,
-            })),
-            material: materials.add(Color::CYAN.into()),
-            transform: Transform::from_translation(Vec3::Y),
-            ..Default::default()
-        });
+    commands.spawn_bundle(PbrBundle {
+        mesh: meshes.add(Mesh::from(shape::UVSphere {
+            radius: 1.0,
+            sectors: 32,
+            stacks: 16,
+        })),
+        material: materials.add(Color::CYAN.into()),
+        transform: Transform::from_translation(Vec3::Y),
+        ..Default::default()
+    });
 
     commands.spawn_bundle(ParticleEffectBundle::new(effect));
 }
--- a/examples/circle.rs
+++ b/examples/circle.rs
@@ -22,7 +22,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .insert_resource(options)
         .insert_resource(bevy::log::LogSettings {
             level: bevy::log::Level::WARN,
-            filter: "bevy_hanabi=trace,circle=trace".to_string(),
+            filter: "bevy_hanabi=error,circle=trace".to_string(),
         })
         .add_plugins(DefaultPlugins)
         .add_plugin(HanabiPlugin)
@@ -57,7 +57,7 @@ fn setup(
             name: "Gradient".to_string(),
             // TODO: Figure out why no particle spawns if this is 1
             capacity: 32768,
-            spawner: Spawner::once(32.0.into()),
+            spawner: Spawner::once(32.0.into(), true),
             ..Default::default()
         }
         .init(PositionCircleModifier {
--- a/examples/circle.rs
+++ b/examples/circle.rs
@@ -57,7 +57,7 @@ fn setup(
             name: "Gradient".to_string(),
             // TODO: Figure out why no particle spawns if this is 1
             capacity: 32768,
-            spawner: Spawner::new(SpawnMode::Once(SpawnCount::Single(32.0))),
+            spawner: Spawner::once(32.0.into()),
             ..Default::default()
         }
         .init(PositionCircleModifier {
--- a/src/spawn.rs
+++ b/src/spawn.rs
@@ -19,6 +19,12 @@ impl<T: Copy> Value<T> {
     }
 }
 
+impl<T: Copy> From<T> for Value<T> {
+    fn from(t: T) -> Self {
+        Self::Single(t)
+    }
+}
+
 /// Spawn rate, in particles per second.
 pub type SpawnRate = Value<f32>;
 
@@ -59,11 +65,22 @@ impl SpawnMode {
 /// Spawner defining how new particles are created.
 #[derive(Debug, Copy, Clone, Serialize, Deserialize)]
 pub struct Spawner {
-    /// Spawn mode.
-    mode: SpawnMode,
+    /// Number of particles to spawn over `spawn_time`
+    num_particles: Value<f32>,
+
+    /// Time over which to spawn `num_particles`, in seconds
+    spawn_time: Value<f32>,
+
+    /// Time between bursts of the particle system, in seconds.
+    /// If this is infinity, there's only one burst.
+    /// If this is `spawn_time`, the system spawns a steady stream of particles.
+    period: Value<f32>,
 
     /// Time since last spawn.
     time: f32,
+
+    /// Sampled value of `spawn_time` until `limit` is reached
+    curr_spawn_time: f32,
 
     /// Time limit until next spawn.
     limit: f32,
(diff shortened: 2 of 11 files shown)


src/spawn.rs
```diff
@@ -1,116 +1,241 @@
...
+    /// Use this, for example, to immediately spawn some particles.
+    pub fn reset(&mut self) {
+        self.time = 0.;
+        self.limit = 0.;
+        self.spawn = 0.5;
+        self.active = true;
+    }
+
+    /// Sets whether the spawner is active.
+    pub fn set_active(&mut self, active: bool) {
```
djeedai, line 119: Maybe need an `is_active()` too?

src/spawn.rs
```diff
@@ -1,116 +1,241 @@
...
+        self.active = active;
+    }
+
+    /// Resamples the spawn time and period.
+    fn resample(&mut self) {
+        self.curr_spawn_time = self.spawn_time.sample();
+        self.limit = self.period.sample();
```
djeedai, line 126: Shouldn't there be a guarantee that `curr_spawn_time <= limit` always? Also we probably need to look at non-constant sampling, which could produce values `<= 0` which I think will break things in `tick()`.

src/spawn.rs
```diff
@@ -1,116 +1,241 @@
...
+
+            let new_time = self.time + dt;
+            if self.time <= self.curr_spawn_time {
+                self.spawn += if self.curr_spawn_time == 0.0 {
+                    self.num_particles.sample()
+                } else if new_time <= self.curr_spawn_time {
+                    self.num_particles.sample() * dt / self.curr_spawn_time
+                } else {
+                    self.num_particles.sample() * (1.0 - self.time / self.curr_spawn_time)
+                };
```
djeedai, line 149: This is `(max(new_time, self.curr_spawn_time) - self.time) / self.curr_spawn_time` no?
- If `max == new_time` then `new_time - self.time == dt` so we're in the first case
- If `max == self.curr_spawn_time` then this is trivially the second case

src/spawn.rs
```diff
@@ -1,116 +1,241 @@
...
+    pub(crate) fn tick(&mut self, mut dt: f32) -> u32 {
+        if !self.active {
+            return 0;
+        }
+
+        // The limit can be reached multiple times, so use a loop
+        loop {
+            if self.limit == 0.0 {
+                self.resample();
+                continue;
```
djeedai, line 138: `new()` needs to `panic!` on `period <= 0` otherwise this is an infinite loop.

src/spawn.rs
```diff
@@ -1,116 +1,241 @@
...
+            }
+
+            let new_time = self.time + dt;
+            if self.time <= self.curr_spawn_time {
+                self.spawn += if self.curr_spawn_time == 0.0 {
```
djeedai, line 143: Maybe be more gentle with numerical precision to avoid a divide by a tiny number below, and instead use this branch for any case where the `curr_spawn_time` is small enough? Like, if `dt` is in the order of magnitude of 10ms, let's say maybe `curr_spawn_time < 1e-4` which is 0.1ms so 1:100 ratio with the typical `dt`.

src/spawn.rs
```diff
@@ -1,116 +1,241 @@
...
+        let count = spawner.tick(1.0);
+        assert_eq!(count, 5);
+    }
+
+    #[test]
+    fn test_rate() {
+        let mut spawner = Spawner::rate(5.0.into());
+        let count = spawner.tick(1.0);
+        assert_eq!(count, 5);
+        let count = spawner.tick(0.4);
```
djeedai, line 207: I'd be more careful and make sure even with numerical imprecision we're past the 40% mark, e.g.
```suggestion
        let count = spawner.tick(0.41);
```
Maybe that works now because of `0.4` but I don't expect that'd work exactly with other values.

src/spawn.rs
```diff
@@ -1,116 +1,241 @@
...
+        assert_eq!(count, 2);
+    }
+
+    #[test]
+    fn test_rate_accumulate() {
+        let mut spawner = Spawner::rate(5.0.into());
+
+        let count = (0..12).map(|_| spawner.tick(1.0 / 60.0)).sum::<u32>();
+        assert_eq!(count, 1);
```
djeedai, line 228: Same, maybe I'd tick a _tiny_ bit more to be sure we passed `count == 1`. Here we're relying quite a lot on the numerical stability above to give exact results.

djeedai (changes requested):
Thanks for that contribution, this is really cool!

I'm not sold yet on the `reset()` changing the active state, I think that's confusing and not really useful; it's more explicit for the user to call `set_active()` directly. Unless there's something I didn't understand, but I don't think _e.g._ burst will get inactive after the burst so there's no reason to mix the active state with the spawner timings, is there?

I also want to test the examples, they look pretty interesting. I'll try to give it some time later tonight.

src/spawn.rs
```diff
@@ -1,116 +1,241 @@
...
+        Self::new(count, 0.0.into(), period)
+    }
+
+    /// Resets and activates the spawner.
+    /// Use this, for example, to immediately spawn some particles.
+    pub fn reset(&mut self) {
+        self.time = 0.;
+        self.limit = 0.;
+        self.spawn = 0.5;
+        self.active = true;
```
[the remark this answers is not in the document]
josh65536: I was trying to think of a way to allow the user to initialize a spawner that spawns a bunch of particles at once, but not immediately, and then call a single function to tell the spawner to spawn particles. If `reset()` doesn't change the state, then they will need to call both `reset()` and `set_active()` to do this (or call `set_active()` the first time and `reset()` later times, but this isn't elegant either).

src/spawn.rs
josh65536: It's actually `min`, but I'll make the change.

src/spawn.rs
josh65536: It works now because I initialized `spawn` to 0.5, which should make it stay clear of the edge case. I could make the change and then also allow the total count to be off by 1 on the high end.

src/spawn.rs
josh65536: Same as above

src/spawn.rs
```diff
@@ -1,116 +1,290 @@
...
+        // The limit can be reached multiple times, so use a loop
+        loop {
+            if self.limit == 0.0 {
+                self.resample();
+                continue;
+            }
+
+            let new_time = self.time + dt;
+            if self.time <= self.curr_spawn_time {
+                self.spawn += if self.curr_spawn_time < dt / 100.0 {
```
djeedai, line 173: This is a bit dangerous, if `dt == 0.0` this doesn't prevent the divide by zero below.

src/spawn.rs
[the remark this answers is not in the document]
djeedai: If the use case is "spawn a burst of particles each time some event occurs" then I'd expect that's a case for `reset()`. And then what we're missing here is not `set_active()`, but rather the ability to start a `Spawner::once()` in a state where it's not bursting when it becomes active. That way the event handler can just do a `reset()` and the spawner can stay active all the time. What about something like:
```rust
pub fn once(count: Value<f32>, start_active: bool) -> Self {
    let mut spawn = Self::new(count, 0.0.into(), f32::INFINITY.into());
    if !start_active {
        mut_spawn.time = 1.0;
    }
    spawn
}
```
Since `self.period` is INF, and so is `self.limit`, then `self.time` never gets reset automatically (only manually via `reset()`), so this makes the condition `self.time <= self.curr_spawn_time` always false and the spawner never spawn on activation.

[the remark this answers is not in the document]
djeedai: Note: `start_active` is a very bad name, as it's unrelated to `set_active()` / `is_active()`.

src/spawn.rs
djeedai: I don't understand, the spawner is spawning 5 particles per second. On first tick it goes from t=0 to t=1, spawning 5, which is already on an edge case (could be 4.99999 and `floor()` would round to 4, failing the assert). Then on second tick it goes from t=1 to t=1.4, again on an edge case spawning (5.0 * 0.4) = 2 particles. No?

src/spawn.rs
```diff
@@ -1,116 +1,241 @@
...
+    ///   If this is infinity, then there is only 1 burst.
+    pub fn new(count: Value<f32>, time: Value<f32>, period: Value<f32>) -> Self {
+        Spawner {
+            num_particles: count,
+            spawn_time: time,
+            period,
+            time: 0.,
+            curr_spawn_time: 0.,
+            limit: 0.,
+            spawn: 0.5,
```
djeedai, line 106: Oh alright I didn't see that earlier. I don't think we should do that. Why not start at 0.0? We want to be precise on the accumulator here, why start with an half-particle already spawned?

examples/circle.rs
```diff
@@ -57,7 +57,7 @@ fn setup(
             name: "Gradient".to_string(),
             // TODO: Figure out why no particle spawns if this is 1
             capacity: 32768,
-            spawner: Spawner::new(SpawnMode::Once(SpawnCount::Single(32.0))),
+            spawner: Spawner::once(32.0.into()),
```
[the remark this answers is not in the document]
josh65536: The example works with the changes I just pushed.

djeedai (approved):

PR #6, opened 2022-03-25
Fix workgroup calculation

djeedai:
Ensure enough workgroups are dispatched to process all particles,
rounding up to the number of particles to avoid truncating the last few
ones.


djeedai:
FYI @josh65536 this is why setting `capacity < 64` doesn't render anything, because it was rounding down to 0 and not dispatching any compute thread.

PR #11, opened 2022-04-05
Pulling force field effect with example 

eliotbo:


examples/force_field.rs
```diff
@@ -0,0 +1,175 @@
...
+            speed: 0.2,
+            dimension: ShapeDimension::Surface,
+            ..Default::default()
+        })
+        .update(bevy_hanabi::PullingForceFieldModifier::new(vec![
+            PullingForceFieldParam {
+                position_or_direction: attractor2_position,
+                max_radius: 1000000.0,
+                min_radius: BALL_RADIUS * 6.0,
+                mass: -1.0,
```
djeedai, line 103: It's a bit awkward to have negative masses, why is that?

examples/force_field.rs
```diff
@@ -0,0 +1,175 @@
...
+        effect.maybe_spawner().unwrap().reset();
+    }
+}
+
+#[derive(Default, Debug)]
+pub struct MousePosition {
+    pub position: Vec2,
+}
+
+fn record_mouse_events_system(
```
djeedai, line 143: I'm surprised there's nothing simpler. Can't we just read [`Window.cursor_position()`](https://docs.rs/bevy/latest/bevy/window/struct.Window.html#method.cursor_position) when the user clicks a button?

src/asset.rs
```diff
@@ -19,6 +22,9 @@ pub struct UpdateLayout {
     /// Constant accelereation to apply to all particles.
     /// Generally used to simulate some kind of gravity.
     pub accel: Vec3,
+    pub force_field: [PullingForceFieldParam; FFNUM],
+    pub box_lower_bounds: Vec3,
```
djeedai, line 26: Can you briefly document those please?

src/asset.rs
```diff
@@ -7,7 +7,10 @@ use bevy::{
 };
 use serde::{Deserialize, Serialize};
 
-use crate::{Gradient, InitModifier, RenderModifier, Spawner, UpdateModifier};
+use crate::{
+    modifiers::PullingForceFieldParam, modifiers::FFNUM, Gradient, InitModifier, RenderModifier,
```
djeedai, line 11: I generally try to stay consistent with the pattern "do not duplicate paths", so `modifiers::{FFNUM, PullingForceFieldParam}` rather than `modifiers::PullingForceFieldParam, modifiers::FFNUM`.

src/modifiers.rs
```diff
@@ -226,3 +230,115 @@ impl UpdateModifier for AccelModifier {
...
+    /// of the constant field.
+    /// [`PullingForceType::Constant`]
+    pub position_or_direction: Vec3,
+    /// For a pulling/repulsing force, this is the maximum radius of the sphere of influence, outside of which
+    /// the force field is null.
+    pub max_radius: f32,
+    /// For a pulling/repulsing force, this is the minimum radius of the sphere of influence, inside of which
+    /// the force field is null, avoiding the singularity at the source position.
+    pub min_radius: f32,
+    /// Mass is proportional to the intensity of the force. Note that the particles_update.wgsl shader will
```
djeedai, line 285: Isn't it rather the opposite, "the intensity of the force is proportional to the mass"? It's a detail though.

examples/force_field.rs
djeedai: I mean, for repulsors it should be negative, but for attractors it should be positive. Since in the case of gravity we should have something like `F = G (m1.m2) / r^2`, so positive attraction force if both masses are positive.

src/render/particles_update.wgsl
```diff
@@ -142,9 +157,99 @@ fn main([[builtin(global_invocation_id)]] global_invocation_id: vec3<u32>) {
...
+        
+        // if the particle is within the min_radius of a source, then forget about
+        // the other sources and only use the conformed field, thus the "* min_dist_check"
+        pulling_force =  pulling_force * min_dist_check + force_component;
+            
+    }
+    ///////////// End of force field computation /////////////
+
+    // delete this when working in 3d
+    pulling_force.z = 0.0;
```
djeedai, line 222: Delete? Or provide a way to make it 2D/planar if possible, but this should be 3D by default, since the library uses the 3D rendering path of Bevy.

src/render/particles_update.wgsl
```diff
@@ -142,9 +157,99 @@ fn main([[builtin(global_invocation_id)]] global_invocation_id: vec3<u32>) {
...
+    // onto a plane that is tangent to the sphere.
+    let eps = vec3<f32>(0.000001);
+    let v1 = vVel - proj(unit_p2p_conformed + eps, vVel + eps);
+
+    let conformed_field = 
+        (1.0 - not_conformed_to_sphere) * normalize(v1) * length(vVel);
+
+    // // Euler integration
+    vVel = 
+        (vVel + (spawner.accel * sim_params.dt)  + (pulling_force * sim_params.dt))
```
djeedai, line 235: So technically it's not `pulling_force` but `pulling_acceleration` right? We'd need to multiply by the particle's own mass here if that was really a force. Fine for now though.

src/render/particles_update.wgsl
```diff
@@ -142,9 +157,99 @@ fn main([[builtin(global_invocation_id)]] global_invocation_id: vec3<u32>) {
         }
     }
 
-    // Euler integration
-    vVel = vVel + (spawner.accel * sim_params.dt);
-    vPos = vPos + (vVel * sim_params.dt);
+
+    ///////////// Start of force field computation /////////////
```
djeedai, line 161: This entire block should be injected when the modifier is used only. See how it's done for the position init code with `{{INIT_POS_VEL}}` above, and the string replace in `specialize()`:

https://github.com/djeedai/bevy_hanabi/blob/e3ef5af4b3ef9fc286f2ca57a8e1975eda7dcf3a/src/render/mod.rs#L429

src/render/particles_update.wgsl
```diff
@@ -142,9 +157,99 @@ fn main([[builtin(global_invocation_id)]] global_invocation_id: vec3<u32>) {
...
+    
+    // project on the sphere if within conforming distance
+    let pos_to_source = conforming_source - vPos ;
+    let difference = length(pos_to_source) - conforming_radius;
+    vPos = vPos  + difference * normalize(pos_to_source ) * (1.0 - not_conformed_to_sphere) ;
+
+
+
+
+    vVel = (vPos - temp_vPos) / sim_params.dt;
```
djeedai, line 252: At the moment, and this is awkward, Bevy will return `Time::delta_seconds() == 0` on first frame. So `dt == 0` happens, and that produces an infinite velocity. I'm afraid that may cause issues.

djeedai (changes requested):

examples/force_field.rs
eliotbo: I must admit that I didn't think about it twice when I wrote the code, but I have now and I recall that when dealing with Coulomb electric forces, the convention is that attractive force is negative because the charges of the two particles are opposite. I really don't mind changing the sign of the mass though, because I do believe users will default to a positive mass and an attractive force. Perhaps the best way to deal with this is having an explicit parameter that is either PointSourceType::Attractive or PointSourceType::Repulsive.

examples/force_field.rs
eliotbo: I was planning to eventually use a PerspectiveCamera with bevy_mod_picking, so that we can see the 3d nature of the effect. Thoughts?

src/render/particles_update.wgsl
eliotbo: I guess it's in a state of ambiguity for the time being, since the particles are without mass. I'll change it though since acceleration would be more future proof, although we will have to be careful if particle mass is ever implemented. I'll leave a comment for posterity.

examples/force_field.rs
eliotbo: bevy_mod_picking might be overkill, but the question about the PerspectiveCamera remains.

examples/force_field.rs
eliotbo: ForceSourceType might be a better name.

djeedai: That makes sense. My guess is that gravity-like forces is more common than Coulomb so users might tend to be more familiar with positive forces for attractors. `ForceSourceType` sounds ok to me, but that would be an enum right? So that means you'd need to verify that both variant only take a positive "absolute" value, and internally set the sign correctly? And can we add a bypass via `From<f32>` for power users who know to set a positive attraction force and negative repulsion one (that might be easier to store/manipulate than an enum).

examples/force_field.rs
djeedai: Yes perspective camera makes sense, only reason it's not there yet is I was being lazy. Ideally if there's a create to control the camera easily to fly around etc. we can use that. `bevy_mod_picking` might be overkill as we don't really "pick" anything, but if that simplifies the examples and avoid littering them with complex 3D projection code that's irrelevant for what the example is trying to demo, then I'm all for it :)

examples/force_field.rs
eliotbo: > and internally set the sign correctly?

yes, that right

Could you explain with code what the bypass would look like? Do you mean having the field `force_sign: f32` in `PullingForceFieldParam`?

At the moment, I've settled for

```
PullingForceFieldParam {
...,
change_force_to_repulsive: bool,
}

impl Into<ForceFieldStd430> for PullingForceFieldParam {
    fn into(self) -> ForceFieldStd430 {
        ForceFieldStd430 {
            ...
            force_sign: if self.change_force_to_repulsive { -1.0 } else { 1.0 },
        }
    }
}
```

let me know, I am not overly attached to this code :)

examples/force_field.rs
eliotbo: I'm using smooth-bevy-cameras for orbiting the camera. I didn't need bevy_mod_picking after all. Turning mouse position into world position in 3d turns out to be really simple if you only have a camera orbiting the origin.

src/modifiers.rs
```diff
@@ -226,3 +230,116 @@ impl UpdateModifier for AccelModifier {
...
+    }
+}
+
+/// Parameters for the components making the force field.
+#[derive(Clone, Copy)]
+pub struct ForceFieldParam {
+    /// The particle_update.wgsl shader interprets this field as a position when the force type is set
+    /// to either [`ForceType::Linear`], [`ForceType::Quadratic`] or [`ForceType::Cubic`].
+    pub position: Vec3,
+    /// For a attracting/repulsing force, this is the maximum radius of the sphere of influence, outside of which
```
djeedai, line 275: > // For a attracting/repulsing force

This is all there is left though, since you removed constant ones, no?

src/modifiers.rs
```diff
@@ -226,3 +230,116 @@ impl UpdateModifier for AccelModifier {
...
+    pub force_field: [ForceFieldParam; FFNUM],
+}
+
+impl ForceFieldModifier {
+    /// Instantiate a ForceFieldModifier.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the number of attractors exceeds [`FFNUM`].
+    pub fn new(point_attractors: Vec<ForceFieldParam>) -> Self {
```
djeedai, line 321: Maybe
```suggestion
    pub fn new(point_attractors: &[ForceFieldParam]) -> Self {
```
is more generic/useable? Ideally even `impl IntoIter<ForceFieldParam>` is the most generic way, since all we need here is an iterator.

src/render/force_field_code.wgsl
```diff
@@ -0,0 +1,83 @@
...
+    // let temp_vPos = vPos;
+    vPos = (vPos + (vVel * sim_params.dt));
+    
+    // project on the sphere if within conforming distance
+    let pos_to_source = conforming_source - vPos ;
+    let difference = length(pos_to_source) - conforming_radius;
+    vPos = vPos  + difference * normalize(pos_to_source ) * (1.0 - not_conformed_to_sphere) ;
+
+    // // commented because of the potential bug where dt could be zero, although the simulation
+    // // works anyways, needs investigation
```
djeedai, line 82: This is likely because the spawner sets a non-zero velocity and it's never touched after, no?

djeedai (changes requested):
This is getting there, I tried the new example it looks super cool, thanks for the hard work!

src/render/force_field_code.wgsl
eliotbo: I really don't know. I checked whether `dt` was zero-valued or not on the first frame, and indeed it was.

src/modifiers.rs
eliotbo: Good idea!

src/render/force_field_code.wgsl
```diff
@@ -0,0 +1,83 @@
...
+                + (1.0 - not_conformed_to_sphere) 
+                * spawner.force_field[kk].min_radius / 1.2
+                * (1.0 - min_dist_check);
+        }
+
+        let point_source_force =             
+            - unit_p2p
+            * min_dist_check * max_dist_check
+            * spawner.force_field[kk].mass / 
+                (0.0000001 + pow(distance, f32(spawner.force_field[kk].force_type)));
```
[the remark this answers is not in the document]
eliotbo: amen to that!

djeedai (approved):

PR #18, opened 2022-04-09
Added 2D camera support

josh65536:
For now, exactly one of the features "2d" and "3d" must be enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,17 @@ The Hanabi particle system is a modern GPU-based particle system for the Bevy ga
 
 The 🎆 Bevy Hanabi plugin is only compatible with Bevy v0.6.
 
+### Add the dependency
+
+Add the `bevy_hanabi` dependency to Cargo.toml:
+
+```
+[dependencies]
+bevy_hanabi = { version = "0.1", features = ["3d"] }
+```
+
+If you're using a 2D camera, add the "2d" feature instead.
+
 ### System setup
 
 Add the `HanabiPlugin` to your app:
@@ -213,6 +224,9 @@ cargo run --example random --features="bevy/bevy_winit"
   - [ ] Generic 3D mesh
   - [ ] Deformation
     - [ ] Velocity (trail)
+  - [x] 3D camera support
+  - [x] 2D camera support
+  - [ ] Mixed 3D camera and 2D camera support
 - Debug
   - [x] GPU debug labels / groups
   - [ ] Debug visualization
--- a/README.md
+++ b/README.md
@@ -21,14 +21,14 @@ The 🎆 Bevy Hanabi plugin is only compatible with Bevy v0.6.
 
 ### Add the dependency
 
-Add the `bevy_hanabi` dependency to Cargo.toml:
+Add the `bevy_hanabi` dependency to `Cargo.toml`:
 
 ```
 [dependencies]
 bevy_hanabi = { version = "0.1", features = ["3d"] }
 ```
 
-If you're using a 2D camera, add the "2d" feature instead.
+If you're using a 2D camera, add the `2d` feature instead.
 
 ### System setup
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -19,6 +19,17 @@ The Hanabi particle system is a modern GPU-based particle system for the Bevy ga
...
+### Add the dependency
+
+Add the `bevy_hanabi` dependency to Cargo.toml:
+
+```
+[dependencies]
+bevy_hanabi = { version = "0.1", features = ["3d"] }
+```
+
+If you're using a 2D camera, add the "2d" feature instead.
```
djeedai, line 31:
```suggestion
If you're using a 2D camera, add the `2d` feature instead.
```

README.md
```diff
@@ -19,6 +19,17 @@ The Hanabi particle system is a modern GPU-based particle system for the Bevy ga
 
 The 🎆 Bevy Hanabi plugin is only compatible with Bevy v0.6.
 
+### Add the dependency
+
+Add the `bevy_hanabi` dependency to Cargo.toml:
```
djeedai, line 24:
```suggestion
Add the `bevy_hanabi` dependency to `Cargo.toml`:
```

README.md
```diff
@@ -213,6 +224,9 @@ cargo run --example random --features="bevy/bevy_winit"
   - [ ] Generic 3D mesh
   - [ ] Deformation
     - [ ] Velocity (trail)
+  - [x] 3D camera support
+  - [x] 2D camera support
+  - [ ] Mixed 3D camera and 2D camera support
```
djeedai, line 229: I'm not sure what that means / if that's even possible?

src/lib.rs
```diff
@@ -101,6 +101,11 @@ pub use plugin::HanabiPlugin;
 pub use render::EffectCacheId;
 pub use spawn::{Spawner, Value};
 
+#[cfg(not(any(feature = "2d", feature = "3d")))]
+compile_error!("Enable either the 2d or 3d feature.");
```
djeedai, line 105:
```suggestion
compile_error!("Enable either the '2d' or '3d' feature.");
```

src/lib.rs
```diff
@@ -101,6 +101,11 @@ pub use plugin::HanabiPlugin;
...
+#[cfg(all(feature = "2d", feature = "3d"))]
+compile_error!("Disable either the 2d or 3d feature. Both at the same time is not supported.");
```
djeedai, line 107:
```suggestion
compile_error!("Disable either the '2d' or '3d' feature. Both at the same time is not supported.");
```

src/render/mod.rs
```diff
@@ -597,6 +600,9 @@ impl SpecializedPipeline for ParticlesRenderPipeline {
                 topology: PrimitiveTopology::TriangleList,
                 strip_index_format: None,
             },
+            #[cfg(feature = "2d")]
+            depth_stencil: None,
```
djeedai, line 604: That's a bit arguable no? Just because it's 2D doesn't mean users don't want a depth buffer, do they? Why this choice?

djeedai (changes requested):
Thanks @josh65536 that's awesome! Some minor changes, mainly the question about disabling depth buffer in 2D. But looks good otherwise!

src/render/mod.rs
djeedai: Ah just saw on Discord this is a Bevy limitation. Let's leave like that for now then. It's a bit surprising though.

README.md
josh65536: I was thinking that we may want to support a game that uses a 3D camera and a 2D camera in different places/times, but now that I think about it, it may be better to just have them double-depend on `bevy_hanabi` with different features.

djeedai (approved):

PR #33, opened 2022-05-30
Lifetime modifier examples

LeonidGrr:


LeonidGrr:
@djeedai Here is obligatory example for lifetime. Not sure if its needed - you mentioned refactoring is ongoing.

djeedai (approved):

PR #34, opened 2022-07-17
Made the InitLayout and UpdateLayout public

werner291:
The `InitModifier` and `UpdateModifier` traits were impossible to implement because the `InitLayout` and `UpdateLayout` struct were private.

This PR makes these structs publicly available. Shouldn't break any existing code that I'm aware of.

### DIFF
--- a/src/asset.rs
+++ b/src/asset.rs
@@ -16,8 +16,11 @@ use crate::{
 /// to define the initial conditions of particles on the GPU.
 #[derive(Default, Clone)]
 pub struct InitLayout {
+    /// Code to define the initial position of particles.
     pub position_code: String,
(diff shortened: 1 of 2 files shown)


werner291:
Please double-check if my doc comments make sense?

What about the `force_field_code` field? It appears to be unused, as far as I can tell.

djeedai:
> Please double-check if my doc comments make sense?
> 
> What about the `force_field_code` field? It appears to be unused, as far as I can tell.

Comments look good, and it looks like you're right about the force field code. I'll have a further look when I'm on my PC see what's going on exactly, thanks for pointing out!

djeedai (approved):

PR #40, opened 2022-08-12
Add BillboardModifier to force particles to render facing the camera

### LINKED ISSUE #38: Particles always render facing along the X-axis
Not sure if this qualifies as an issue, but I'm not sure how to get the particle system to work the way I'd like it to, and I'd really appreciate some help if you can spare the time :)

I'm using Bevy 0.8.0 and Hanabi 0.3.0 to simulate space dust in a 3D game, but the particles always render facing down the X-axis, and I'm not sure how to go about correcting them. I assumed the particles would always face the camera, but that doesn't appear to be the case.

When viewing the particle effect cluster from along the X-axis, it looks perfect:
Screenshot from 2022-08-11 10-46-46

When viewing from the Z-axis, the particles are barely visible, because they're perpendicular to the viewing angle:
Screenshot from 2022-08-11 10-49-06

I initially just tried to create two perpendicular particle effects to circumvent the issue, but the rotation component of the transform is also ignored by the renderer, which makes sense in hindsight, but I had to try :)

The code that I'm using to create the clouds is [here](https://github.com/Bromidity/star-lost/blob/05d19657044fee5b396ae54a9d4947e117f51da6/src/dust.rs)
My camera definition, which I suspect might be relevant is defined [here](https://github.com/Bromidity/star-lost/blob/05d19657044fee5b396ae54a9d4947e117f51da6/src/tests/controls.rs#L31-L35). I'm parenting it to a "Ship" entity, which I suspect might be relevant.

Anyway, thanks for creating this plugin, even with this issue, the particle system is amazing and helped me get over a mental block!

MathiasPius:
I think I've managed to get some version of the feature working, although I'm not sure if this is how you imagined it. Implementing it during the render phase seemed the easiest, but I could imagine a scenario where being able to manipulate or rotate the particles during the update phase could be useful, which would probably conflict with the work done here.

This is a pretty simplistic implementation that doesn't support looking in arbitrary directions, only directly at the camera. Let me know what you think!

Fixes #38


djeedai (approved):
Fantastic, thanks a lot for the contribution! There's probably room for improvement, but I see no reason not to merge as is for now, especially as you took the time to write an example for it. Again, thanks for the work, much appreciated!

MathiasPius:
My pleasure! Thanks for doing the other 99% of the work that let me put space dust in my game :smile:

PR #57, opened 2022-10-02
Fix item alignment for GPU in `AlignedBufferVec`

### LINKED ISSUE #55: bound buffer range 0..592 does not fit in buffer of size 584
After updating to latest `main` branch, I get the following error when creating a spawner:

```
thread 'Compute Task Pool (5)' panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_bind_group
      note: label = `hanabi:spawner_bind_group`
    bound buffer range 0..592 does not fit in buffer of size 584
      note: buffer = `<Buffer-(64, 421, Vulkan)>`
```

djeedai:
Ensure that `AlignedBufferVec` items are properly aligned according to the WGSL language rules.

Fixes #55

### DIFF
--- a/src/render/aligned_buffer_vec.rs
+++ b/src/render/aligned_buffer_vec.rs
@@ -1,8 +1,12 @@
+use std::num::NonZeroU64;
+
 use bevy::{
-    core::{cast_slice, Pod},
+    core::{cast_slice, Pod, Zeroable},
     log::trace,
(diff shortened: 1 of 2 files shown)


djeedai:
Reported Rust nightly compiler bug, waiting a day or two to see if that's fixed and we can have a CI-validated change.

PR #88, opened 2022-11-13
Upgrade to Bevy 0.9

djeedai:


src/asset.rs
```diff
@@ -66,7 +66,7 @@ pub struct RenderLayout {
 ///
 /// [`ParticleEffect`]: crate::ParticleEffect
 /// [`ParticleEffectBundle`]: crate::ParticleEffectBundle
-#[derive(Default, Clone, Serialize, Deserialize, Reflect, TypeUuid)]
+#[derive(Default, Clone, Serialize, Deserialize, Resource, Reflect, TypeUuid)]
 #[reflect(Resource)]
```
haihala, line 70: Is this necessary anymore?

src/render/shader_cache.rs
```diff
@@ -1,6 +1,6 @@
 use bevy::{
     asset::{Assets, Handle},
-    ecs::change_detection::ResMut,
+    ecs::{change_detection::ResMut, system::Resource},
```
haihala, line 3: A lot of these can be imported from prelude.

haihala (approved):
Not an experienced rust dev by any stretch of the imagination, but looks good to me.

src/asset.rs
djeedai: It doesn't build without it. Why do you expect it's not necessary? Bevy 0.9 makes `Resource` derive explicit.

src/render/shader_cache.rs
djeedai: I'm conflicted about the use of `prelude` in library crates to be honest. It's probably a bias from C/C++, and likely an incorrect one, but I tend to get worried about importing large groups of things that I don't need. Especially for Bevy, where the prelude is huge, and there are known bugs with `rustdoc` getting confused with it (but I think they're more on the multi-crate side of things). I'd be happy to review if there's any guidance though.

djeedai: I found [this Reddit discussion](https://www.reddit.com/r/rust/comments/q6stlt/massive_preludes_why/) were people seem more against preludes, although there's no clear cut nor really convincing argument.

src/asset.rs
haihala: The marking in way less clear in this view, I'm trying to refer to the `#[reflect(Resource)]` line, since maybe it's redundant after the derive. Could not be the case.

djeedai: I will leave out for now, I've never understood what those `#[reflect(xxx)]` do that `#[derive(Reflect)]` cannot do automatically.

Issue #95, opened 2022-11-30
Cone distribution not random


NiseVoid:
When creating a cone, particle distribution is not uniform. I think the problem comes from alpha_h being set to `pow(rand(), 1.0/3.0)`. This is not correct and performs even worse than setting it directly to `rand()`, which would at least behave correctly on a cylinder (top_radius == bottom_radius) and looks somewhat uniform on any cone without a sharp point.

djeedai:
Could you back those claims with resources or mathematical proofs? The 1/3 factor is derived mathematically from a half-cone. Maybe it doesn't extend to a truncated cone, I can't remember the details to be honest, but I find it hard to believe a simple `rand()` would perform any better in the general case.

NiseVoid:
After I changed some code so top_radius is the side facing Y+ it works fine for perfect cones. However, any cone without `top_radius: 0` will not have uniform distribution, be it reversed (`top_radius > base_radius`), non-perfect (`top_radius != 0`), or a cylinder (`top_radius == base_radius`). `alpha_h` would need to be calculated using these radii, to handle all these cases correctly.

djeedai:
This is the distribution for a cone of height `10.` with a top of `0.` and base of `10.`, so a perfect (half) cone (truncated at the base). The white cube indicates the origin of the effect, at the top/apex of the cone.

![image](https://user-images.githubusercontent.com/6532395/204917926-48514bff-8699-426e-84d2-7de801473a8a.png)

Distribution looks fairly uniform.

Now with top radius `5.`:

![image](https://user-images.githubusercontent.com/6532395/204917986-e18a1f4f-8d45-4d77-b905-8fa4fc37897b.png)

We can start to guess some issue.

With top radius `10.`, forming a cylinder, we finally see the issue clearly:

![image](https://user-images.githubusercontent.com/6532395/204918087-fc00e5ec-c721-4ca5-8e5f-d47c1e477ec8.png)

Note that there's no "change the facing" involved here. The origin is at the top by design, which is the apex when `r = 0`, and the base is always the larger side. This goes hand in hand with the face particles are emitted at the base toward +Y, which we can infer from the colors when setting a non-zero speed:

![image](https://user-images.githubusercontent.com/6532395/204918431-ad6cf73e-7c89-4a7a-9ca1-5b8b247ff20b.png)

[...]

PR #96, opened 2022-11-30
Make top_radius the top for PositionCone3dModifier

NiseVoid:
I assume the top_radius is meant to be at the higher Y coordinate, this PR makes it so that that's true.

### DIFF
--- a/src/modifier/init.rs
+++ b/src/modifier/init.rs
@@ -195,15 +195,15 @@ impl InitModifier for PositionCone3dModifier {
     // Truncated cone height
     let h0 = {0};
     // Random height ratio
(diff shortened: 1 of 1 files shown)


NiseVoid:
```
.init(PositionCone3dModifier {
    height: 30.,
    top_radius: 0.,
    base_radius: 5.,
    dimension: ShapeDimension::Volume,
    speed: Value::Single(0.),
})
```
now results in:

NiseVoid:
As discussed in #95, a change to the fieldname `top_radius` is probably the better solution to the confusion around `top_radius` facing down.

PR #145, opened 2023-02-25
Use less restrictive bytemuck dependency

### LINKED ISSUE #144: Pinned bytemuck
Per https://github.com/djeedai/bevy_hanabi/blob/main/Cargo.toml#L25, `bytemuck` is pinned at `"=1.12.3"`.
This introduces quite a few dependency clashes in my projects, since `bytemuck` is a very popular crate. I see that https://github.com/djeedai/bevy_hanabi/commit/47bff1256285eb297ff7a9f0dad3f5e11b3d485e introduced the pin, but I don't understand why it's needed. 
I ran the `instancing` example mentioned in the PR with both the pinned dependency and `bytemuck` at `1.13.0`, but I found them to behave identically. I did however not run all examples to verify.
Maybe the reason `bytemuck` was pinned no longer applies and we can update the dependency? If not: can I somehow help?

janhohenheim:
Fix #144

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ default = ["2d", "3d", "gpu_tests"]
 gpu_tests = []
 
 [dependencies]
-bytemuck = { version = "=1.12.3", features = ["derive"] }
+bytemuck = { version = "1.13", features = ["derive"] }
 copyless = "0.1"
 rand = "0.8"
 rand_pcg = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ default = ["2d", "3d", "gpu_tests"]
 gpu_tests = []
 
 [dependencies]
-bytemuck = { version = "1.13", features = ["derive"] }
+bytemuck = { version = "1.7.2", features = ["derive"] }
 copyless = "0.1"
 rand = "0.8"
 rand_pcg = "0.3"


djeedai (commented):
Looks like Bevy only uses `1.5` (so, not pinned). Can we upgrade to the same? I generally try to follow Bevy's deps unless there's a reason to diverge.

janhohenheim:
image
Looks like the oldest we could do is `1.7.2`. 
If you don't want to upgrade, I would be quite happy already if we just removed the `=` from the version spec.

djeedai:
I meant `1.5` not `=1.5`. Seems it conflicts because you pinned it no? Because `1.5` should allow cargo to select `1.7.2` or anything `1.x` really. `1.7` is fine too, some Bevy packages use that.

janhohenheim:
I understood that and see where you are coming from. The pin was just to check if "1.5" was sound, since AFAIK writing it like that into your Cargo.toml tells your consumers "I guarantee that my software works with version 1.5 and above". Since we fail to compile with anything below "1.7.2", I argue that writing down "1.5" would break semver guarantees, so I advise against it.
It's your call though, feel free to disagree and I'll set it back to "1.5". Maybe I also understood something wrong about semver, I'm no expert 🙂

djeedai:
I see. Yes your probably right. I'm fine with `1.7.2`.

janhohenheim:
Done :)

djeedai (approved):

Issue #147, opened 2023-03-01
Update to Bevy 0.10


janhohenheim:
Seeing as Bevy 0.10 is about to be released, I thought it's time to put this in the room 🙂

djeedai:
This is blocked by `bevy-inspector-egui` anyway, but should land usually a few days after the Bevy release.

PR #148, opened 2023-03-02
Update to bevy 0.10

NiseVoid:
**The update is blocked by bevy-inspector-egui, but not really since it only affects examples and can easily be disabled**

To run examples: Disable inspector in the Cargo toml and the example code

I only ran examples and tests, they all seemed to work fine, but I'm not sure if any features have been missed in these. I did have some weird errors when closing windows, and some of the tests seemed to eat a lot of GPU resources, but I'm not sure if this is new behavior.

Some things to note:

- Shaders now need to specify global variables with `const` instead of `let`.
- You can no longer redefine a variable name with `let`.
- The set EffectSystems::ExtractEffects is unused, and only lives in the ExtractSchedule, so it can probably be removed
- There is a new field on `RenderPipelineDescriptor` called `push_constant_ranges`, I left it empty but there might be a better value.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,12 @@ wgpu = "0.15"
 naga = "0.11"
 
 futures = "0.3"
-bevy-inspector-egui = "0.17"
+bevy-inspector-egui = "0.18"
+
+# Enable default features for examples and tests
+[dev-dependencies.bevy]
+version = "0.10"
+default-features = true
 
 [[example]]
 name = "firework"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,8 +33,7 @@ bitflags = "1.3"
 typetag = "0.2"
 
 [dependencies.bevy]
-git = "https://github.com/bevyengine/bevy"
-branch = "main"
+version = "0.10"
 default-features = false
 features = [ "bevy_core_pipeline", "bevy_render", "bevy_asset" ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,8 @@ bitflags = "1.3"
 typetag = "0.2"
 
 [dependencies.bevy]
-version = "0.9"
+git = "https://github.com/bevyengine/bevy"
+branch = "main"
 default-features = false
 features = [ "bevy_core_pipeline", "bevy_render", "bevy_asset" ]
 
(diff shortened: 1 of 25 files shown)


djeedai:
Thanks for the preparatory work @NiseVoid. That's going to be very useful for a swift upgrade :) I'm going to wait the last minute to merge though, just in case there would be any need to release on v0.9 first (there was a lot of changes since v0.5). Better safe than reverted. But I do intend to release as soon as possible after Bevy v0.10 landed.

What I meant by "The update is blocked by bevy-inspector-egui" is that the examples (and therefore CI) can only pass once that crate upgraded. And v0.6 can only be published to crates.io after that, which is really what most folks are waiting for. I once did disable the dependency in examples and released without it, then re-enabled and released again; the result was a "gain" of 24h~48h earlier release over what would have happened if not disabled, but that gain was largely offset by the absolute pain of having to manually do 2 releases back-to-back, with all the overhead (including manual testing) involved. I wish that was not the case and I could just press a button, but because of the difficulty to test on GPU and lack of time to write comprehensive tests, we're unfortunately not quite there yet. So I'm unlikely to do that again, baring exceptional circumstances, and especially since Jakob is generally very diligent and timely with their update of `bevy-inspector-egui`.

janhohenheim:
This closes #147

djeedai:
@NiseVoid will you update that PR once `bevy-inspector-egui` is updated, or should I take over the upgrade? I don't mind either.

NiseVoid:
> @NiseVoid will you update that PR once `bevy-inspector-egui` is updated, or should I take over the upgrade? I don't mind either.

If that's the only issue with the PR, I guess it might be faster if you do that update and can then merge it when done.

I also tested this PR on my code, and it seems to behave identical to the commit it was based on.

djeedai:
> If that's the only issue with the PR, I guess it might be faster if you do that update and can then merge it when done.

That needs to be done as part of this PR though.

Also we should expect an updated `bevy-inspector-egui` for 0.10 within the hour.

NiseVoid:
Updated the inspector, rebased, and ran all examples (without any changes to the Cargo.toml or tests compared to what was comitted). Seems to all work fine

djeedai (approved):

PR #154, opened 2023-03-05
Fixed PositionCone3dModifier placing particles at twice the translation

### LINKED ISSUE #152: PositionCone3dModifier generates particles at wrong location
When using the PositionCone3d the particles are generated at a location that seems to be 2x (in each dimension) the actual translation of the entity.

When going back to the InitPositionCircleModifier the particles are generated around the correct center.

Version: main (specifically 8cbfa36080)

OleStrohm:
The position inside the cone was affected by the transform, but since the w component of the position was set to 1 the transform was affecting adding the translation. However this also gets done when it's rendered I believe, so the translation was added twice.

However I'm not sure why the position (and the speed in the next part) were affected by the transform at all. The code for InitPositionCircleModifier does not use transform, though the InitVelocityCircleModifier does, so maybe this change is the most correct? I can't seem to see any observable difference at all whether I keep it or remove it entirely, but maybe I'm missing something. Though I think maybe neither of them should be affected by the transform?
But maybe I'm missing something here.

Fixes #152

### DIFF
--- a/src/modifier/init.rs
+++ b/src/modifier/init.rs
@@ -245,8 +245,7 @@ impl InitModifier for PositionCone3dModifier {
     let y = h;
     let z = r * sint;
     let p = vec3<f32>(x, y, z);
-    let p2 = transform * vec4<f32>(p, 1.0);
-    (*particle).{3} = p2.xyz;
+    (*particle).{3} = p.xyz;
     // Emit direction
     let rb2 = rb * alpha_r;
     let pb = vec3<f32>(rb2 * cost, h0, rb2 * sint);
--- a/src/modifier/init.rs
+++ b/src/modifier/init.rs
@@ -245,7 +245,8 @@ impl InitModifier for PositionCone3dModifier {
     let y = h;
     let z = r * sint;
     let p = vec3<f32>(x, y, z);
-    (*particle).{3} = p.xyz;
+    let p2 = transform * vec4<f32>(p, 0.0);
+    (*particle).{3} = p2.xyz;
     // Emit direction
     let rb2 = rb * alpha_r;
     let pb = vec3<f32>(rb2 * cost, h0, rb2 * sint);


djeedai (commented):
I think it's important to keep the transform, otherwise the orientation of the emitter won't be taken into account anymore. I think the error is that the translation part of the transform is applied a second time unconditionally after the modifiers. Maybe all modifiers should apply the transform themselves. Or maybe there should be some variable telling the subsequent shader code whether to skip that offset.

https://github.com/djeedai/bevy_hanabi/blob/8cbfa36080709ac43c9bff84edd4869bcabff1f5/src/render/vfx_init.wgsl#L160

OleStrohm:
If you think the orientation is important (I couldn't see it in my game, and I figured the orientation would be taken care of when rendering, since the translation is, the rotation should be too), but if you think that's not handled elsewhere this can be changed to only apply the orientation similarly to what the velocity does, by setting w=0, but I'm not quite sure how this fits into the rest of the code.

Isn't orientation and translation and such handled when rendering?

djeedai:
> Isn't orientation and translation and such handled when rendering?

The position and orientation are attributes of the particle, which are simulated during the update phase. During rendering, we make use of that position and orientation to produce some visuals; currently this means creating a quad centered at the given particle position and oriented by default alongside that particle orientation (XY if I remember correctly, or maybe was it XZ?), unless a render modifier tells us to do something else (like, `BillboardModifier` forcing the quad to face the camera).

So yes and no. Yes rendering does re-orient quads if some modifiers are present (and in that case only; otherwise it just uses the particle's orientation). But no, the particle orientation still need to be properly simulated during the update phase for rendering to make use of it.

> by setting w=0

I guess we _can_ do it like that yes, although I think that doesn't really solve the larger issue that both the cone modifier and the shader template apply (part of) the `Transform` of the emitter, and that's a bit confusing. But we can revisit later, as long as we fix for now the bug.

djeedai:
Actually sorry a few mistakes:

1. The removal of `let p2 = transform * vec4<f32>(p, 1.0);` has nothing to do with rotation, but with scale. Using instead `let p2 = transform * vec4<f32>(p, 0.0);` would apply any eventual scale, while removing it will not.
2. The rotation is already handled by `let dir = transform * vec4<f32>(normalize(pb - p), 0.0);` below.

OleStrohm:
I just did a proper deep dive, and it turns out the speed code had been confusing me, but when I turned it off it was obviously way easier to see the effects on the position part.
I've now confirmed that multiplying by the transform (with w=0) leads to the correct orientation and scale and translation, so you are indeed right. I'll update this PR now.
Sorry for the confusion!

BorisBoutillier:
I encountered this issue in migrated Kataster to use 0.6 version of bevy_hanabi.
The cone3d particle effects that I use to represent thrust exhaust was not behaving correctly on modification of its transform.

I have tested with the modification in the PR and the cone of particle now behaves itself appropriately.

BorisBoutillier (approved):
Check this fix on my problematic use case. The cone of particle now behave correctly on translation and rotation changes.

djeedai:
Yes I moved forward with the release to not block other folks but I'll do a 0.6.1 with that fix when merged.

djeedai (approved):

PR #174, opened 2023-03-30
Add `starts_immediately` related methods

zArubaru:
This change makes it possible to create a single burst of particles over a time window, that doesn't fire immediately. My workaround before was to have it inactive at first, but this seemed counter-intuitive, as we have the `once` function.

Perhaps an additional function we could add, is something like `burst_once_over_time`, pending for a better name 😅


djeedai:
Hi- Thanks for the PR!

> This change makes it possible to create a single burst of particles over a time window, that doesn't fire immediately.

How is that different from the `spawn_on_command.rs` example which also uses a `once` spawner and `reset()` to trigger the spawning when necessary? I'm not necessarily opposed to that API but I'd like to understand what's the use case.

zArubaru:
Hey, thanks for getting back to me!

I added an example, see it [here](https://github.com/zArubaru/bevy_hanabi/tree/feature/spawner-time-manipulation#burst-over-time-on-command). Scroll back to the previous example ([spawn_on_command](https://github.com/zArubaru/bevy_hanabi/tree/feature/spawner-time-manipulation#spawn-on-command)) to see the difference.

I may also be missing something, but as far as I could tell, with the current API in `main` I could only get this to work with having it deactive on start.

examples/burst_over_time_on_command.rs
```diff
@@ -0,0 +1,178 @@
...
+            }),
+            ..Default::default()
+        })
+        .insert(Ball {
+            velocity: Vec2::new(1.0, 2f32.sqrt()),
+        })
+        .insert(Name::new("ball"))
+        .id();
+
+    let spawner = Spawner::new(32.0.into(), 0.5.into(), std::f32::INFINITY.into()).with_time(0.5);
```
djeedai, line 101: nit: I'd set an initial time > 0.5 to avoid confusion, and maybe add a comment:
```suggestion
    // Set an initial time of 1.0s greater than the burst duration 0.5s to prevent spawning on start.
    let spawner = Spawner::new(32.0.into(), 0.5.into(), std::f32::INFINITY.into()).with_time(1.0);
```

src/spawn.rs
```diff
@@ -348,6 +348,44 @@ impl Spawner {
...
+    ///
+    /// ```
+    /// # use bevy_hanabi::Spawner;
+    /// // Ready a spawner that fires 32 particles in a burst over 1 second after [`reset()`].
+    /// let mut spawner = Spawner::new(32.0.into(), 1.0.into(), std::f32::INFINITY.into());
+    /// spawner.set_time(1.0);
+    /// ```
+    ///
+    /// [`reset()`]: crate::Spawner::reset
+    pub fn set_time(&mut self, time: f32) {
```
djeedai, line 380: This is not going to work after #176 merged, because the current (runtime) time moved to `EffectSpawner`. Also I guess we might want to add a `Spawner::initial_time` to initialize `EffectSpawner::time`, instead of the `EffectSpawner::starts_immediately` present in #176.

djeedai (changes requested):
That looks like a useful addition, that will probably be cleaner that what I did in #176, however #176 moves some things around so this needs to be rebased.

zArubaru:
@djeedai I made some changes, dropped `with_time` and added `starts_immediately` related functionality. See the latest commit :). I tested that all the examples work correctly (and updated my new example also). What do you think?

djeedai (commented):
I'm still not convinced, sorry 😅
- The added methods for `starts_immediately` are probably fine.
- The time ones are... maybe ok
- The fact that the infinite time has a special effect though, that one I'd like to be more convinced that I am now that it's a good thing. See other comment for details.

djeedai:
> Also not sure do we really need get_time and set_time but I can imagine there can be some cases where they are useful if someone needs temporal precision 😅🤔

It's always good to expose this kind of things, both for testing and for tooling (some future VFX editor), even if not used much at runtime in games/apps, provided it doesn't force any design or behavior that is detrimental to runtime usability/performance. I'd probably keep those.

zaqxsw-dev:
> Hey, thanks for getting back to me!
> 
> I added an example, see it [here](https://github.com/zArubaru/bevy_hanabi/tree/feature/spawner-time-manipulation#burst-over-time-on-command).
> 
> I may also be missing something, but as far as I could tell, with the current API in `main` I could only get this to work with having it deactive on start.
> 
> As a direct comparison how it's different to `spawn_on_command`:
> 
> spawn_on_command [ spawn_on_command ](https://raw.githubusercontent.com/djeedai/bevy_hanabi/ffbf91be7f0780f8830869d14a64a79ca79baebb/examples/spawn_on_command.gif) [ ](https://raw.githubusercontent.com/djeedai/bevy_hanabi/ffbf91be7f0780f8830869d14a64a79ca79baebb/examples/spawn_on_command.gif)
> 
> ... and `burst_over_time_on_command` (the example in the README.md is intentionally more different):
> 
> burst_over_time_on_command [ burst_over_time_on_command ](https://raw.githubusercontent.com/zArubaru/bevy_hanabi/42ae87567d6ec8adac79a3e7f092fd7ff8b62b09/examples/burst_over_time_on_command.gif) [ ](https://raw.githubusercontent.com/zArubaru/bevy_hanabi/42ae87567d6ec8adac79a3e7f092fd7ff8b62b09/examples/burst_over_time_on_command.gif)

I think you can do it in current api using Spawner or add this new function in Spawner implementation

`pub fn until(count: Value<f32>, spawn_immediately: bool, timeout: f32) -> Self {
        let mut spawner = Self::new(count, 0.0.into(), timeout);
        spawner.starts_immediately = spawn_immediately;
        spawner
    }`

djeedai:
I've just pushed a change to clean-up the spawner (now `SpawnerSettings`). Spawning a "burst over time" is already supported by using the `new()` constructor, although you have to set the individual values yourself. We could add a helper for this, but I'm concerned because thinking more about your example where you modified `spawn_on_command` it's deeply flawed : because you're using a single effect, each new burst will interrupt the previous one. You can see that when the ball hits the lower right corner : there's a single green particle emitted, because the burst is immediately interrupted by a second collision which moves the emitter and starts a new burst. This means that the number of particles spawned depends on how much time there is until the next spawn event, and it's _highly_ unlikely that's what anyone intended to author.

Typically to achieve this feature of reacting to game events like physics collisions, other users have used a pool of effect instances, fetching an instance from the pool when there's a collision, and recycling it once the burst is done. The pool doesn't need to be large, for short effects like this you just have to evaluate how many bursts you expect to play at the same time. So maybe just 10 or less, depends on use cases. This way there's no timing issue, and all bursts can play integrally.

So I'm going to close this PR. I feel like a helper would encourage that pattern and help users shoot themselves in the foot. If you feel there's a strong need to add some helper nonetheless or other, let's discuss first on a GitHub issue and settle on a solution. Thanks!

Issue #41, opened 2022-08-16
WASM compatibility


ZettaScript:
Using `bevy_hanabi` fails in WASM because `VERTEX_WRITABLE_STORAGE` is not supported (even if WGPU doc says it is supported by all platforms).

Is it planned to make this crate WASM-compatible?

djeedai:
No, `bevy_hanabi` uses compute shaders which are not currently supported on Web.

djeedai:
I'm going to close this issue. I fully appreciate the value of wasm, and I'm kind of sad that this crate doesn't work with it, but it would be such a large amount of work not only to create a separate backend dedicated to wasm that doesn't use compute shaders, but also to maintain two backends in parallel, that I realistically don't have the time to do this.

I hear some Bevy users are working on a CPU particle solution, which would work with wasm (see bevyengine/rfcs#28) so I encourage anyone who needs a particle system that works in wasm to contribute to that RFC and workstream.

If you're familiar with Unity, and to hazard an over-flattering comparison that is probably not realistic (yet?), 🎆 Bevy Hanabi is to Unity's VFX Graph what that CPU solution and RFC for Bevy is to Unity's built-in particle system, with the same advantages and drawbacks. CPU solutions are more versatile, easier to support on all platforms, but limited in terms of performance; on the other hand GPU solutions are vastly more performant but require modern API and hardware features (compute shaders) which are not available everywhere.

johanhelsing:
too, bad, but I totally undestand your decision. I guess we can only hope that webgpu will solve this in the future...

maybe add a small note in the readme that wasm is not supported?

djeedai:
> maybe add a small note in the readme that wasm is not supported?

I thought I already had but apparently not. Fixed now, it's in the README.

janhohenheim:
AFAIK compute shaders are now supported by WebGPU. @djeedai would you consider reopening this issue?

djeedai:
@janhohenheim I would be more than happy to add wasm support to 🎆 Hanabi provided Bevy itself supports compute shaders on wasm. I don't believe this is the case yet though? And I can really spearhead that myself.

janhohenheim:
@djeedai hmm, I tried the `compute_shader_game_of_life` example on main and forced the wgpu backend to WebGPU by calling the following on the `DefaultPlugin` setup:
```rust
.set(RenderPlugin {
    wgpu_settings: WgpuSettings {
[...]

PR #187, opened 2023-06-11
Add hyperlink for wasm issue in README

Piturnah:
Very small change, should be redundant soon enough anyway when 0.11 is released


djeedai (approved):
Thanks! I thought GitHub was auto-linking issues? Or did that break?

Piturnah:
> Thanks! I thought GitHub was auto-linking issues? Or did that break?

My understanding is that they do autolink issues and PRs, but not in the context of a render of a markdown file in the repo. Not really sure why this is, likely because it is more often in that context that you might just want to write literally `#41` for example.

PR #189, opened 2023-06-20
Add rand functions to render shader

yrns:
Both the set size and color render modifiers use these random functions which are missing, I believe. None of the examples use the uniform variant for these modifiers. Bevy supports shader imports now, maybe that should be used here?


djeedai (approved):
Yes I will have a look at shader imports but I'm rewriting a good chunk of the modifier API to allow customizing via expression composition so I'll do that after.

PR #190, opened 2023-06-22
Fix attribute name for InitSizeModifier w/ D2

yrns:
It should be "size2" for D2?


djeedai (approved):
Yes good catch, thanks!

PR #191, opened 2023-06-23
Graph API

djeedai:
This change adds a large part of the Graph API, including most of the low-level
Expression API for code-first runtime-focused customizing of modifiers, and an
experimental preview of the higher level editing-focused Node API.

The `graph` module, as well as its `expr` and `node` submodules, contain design
notes and high-level introductions to the new Expression and Node APIs,
respectively.

The CHANGELOG has been updated with an exhaustive list of changes. We describe
here the main changes only.

Notably, this change removes the `InitAgeModifier` and `InitLifetimeModifier`,
which are superseeded by the more generic and powerful `InitAttributeModifier`,
which can initialize any particle attribute. The CHANGELOG lists all modifiers
which have been otherwise impacted by some changes.

This change is a major breaking change, and is accompanied by a migration guide
(in the `docs/` folder) due to the large amount and complexity of changes
compared to previous releases. The migration from v0.6 is relatively
straightforward however, following the guide or looking at the examples.

All examples have been converted to this new Expression API, and are fully
functional. The change should be on parity with the previous architecture;
however if you notice any discrepancy or regression please open a bug.


djeedai:
Comments from @NiseVoid on Discord:

> Some quick findings: It works, but
> -  It's easy to forget to add a module during the migration (because default screws you over)
> -  It feels weird that only a small fraction of fields use this new format (but I guess that might just be future work)
> -  I do see some code for some operators like normalize that don't seem to be available on ExprWriter
> -  Some examples of more complicated effects would be nice, like a reimplementation of radial or tangential accel with just AccelModifier
> -  The change in the README doesn't seem valid

PR #232, opened 2023-10-01
Fix matrix to wgsl string conversion

Bendzae:
I noticed a bug in the `to_wgsl_string` impl for Matrices, where the first element was added twice. I also introduced some basic `From` implementations for the most common matrix formats to be able to test & use matrix values.

Changes:
- Fixed bug in `MatrixValue.to_wgsl_string()`
- Introduced some test cases to verify the fix
- Added `From` implementations for `Mat2`, `Mat3` and `Mat4`


djeedai:
Thanks for the fix! Can you please put a line in the `CHANGELOG.md` about it?

Bendzae:
> Thanks for the fix! Can you please put a line in the `CHANGELOG.md` about it?

Done :)

djeedai (approved):

PR #237, opened 2023-10-19
Introduce "VARIABLE" expression to allow multiple uses of rng results

Bendzae:


src/lib.rs
```diff
@@ -760,8 +760,17 @@ impl EffectShaderSource {
...
+            let var_expressions = module.get_variable_expressions();
             let mut init_context =
                 InitContext::new(&mut module, &property_layout, &particle_layout);
+
+            // Generate code for variables
+            let mut variables_code = String::new();
+            for (name, expr) in var_expressions.iter() {
+                let expr_code = init_context.expr(*expr).unwrap().eval(&init_context).unwrap();
+                variables_code += format!("\nlet {name} = {expr_code};").as_str();
+            }
```
djeedai, line 772: I think this is not going to work because this generates the code for all variables upfront at the top of the function, but some of those expressions may depend on each other or on code generated later.

I'm looking at a different approach for this, which doesn't require manual variable expressions, but instead marks expressions as having a side-effect, and if they do, generate a local variable to store their evaluation before reusing it. This also requires an expression evaluation cache to avoid re-evaluating an expression twice. This seems to work on a simple example `r + r` where `r` is a `Rand` expression, the test I wrote does generate `(var0) + (var0)` (note the same name used twice).

PR #248, opened 2023-11-05
Bevy 0.12

haihala:
Still a work in progress, feedback wanted.

### DIFF
--- a/src/render/mod.rs
+++ b/src/render/mod.rs
@@ -1,7 +1,7 @@
 #[cfg(feature = "2d")]
 use bevy::utils::FloatOrd;
 use bevy::{
-    asset::{AssetEvent, Assets, Handle, HandleId},
+    asset::{AssetEvent, Assets, Handle},
     core::{Pod, Zeroable},
     ecs::{
         prelude::*,
@@ -18,14 +18,14 @@ use bevy::{
         renderer::{RenderContext, RenderDevice, RenderQueue},
         texture::{BevyDefault, Image},
         view::{
-            ComputedVisibility, ExtractedView, Msaa, ViewTarget, ViewUniform, ViewUniformOffset,
-            ViewUniforms, VisibleEntities,
+            ExtractedView, InheritedVisibility, Msaa, ViewTarget, ViewUniform, ViewUniformOffset,
+            ViewUniforms, ViewVisibility, VisibleEntities,
         },
         Extract,
     },
     time::Time,
     transform::components::GlobalTransform,
-    utils::{HashMap, Uuid},
+    utils::HashMap,
 };
 use bitflags::bitflags;
 use rand::random;
@@ -883,7 +883,7 @@ pub(crate) struct ParticleRenderPipelineKey {
 impl Default for ParticleRenderPipelineKey {
     fn default() -> Self {
         Self {
-            shader: Handle::weak(HandleId::new(Uuid::nil(), u64::MAX)),
+            shader: Handle::weak_from_u128(u128::MAX),
             particle_layout: ParticleLayout::empty(),
             has_image: false,
             screen_space_size: false,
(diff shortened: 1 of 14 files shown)


src/test_utils.rs
```diff
@@ -206,41 +205,37 @@ impl MockRenderer {
     }
 }
 
-/// A dummy asset IO, just to satisfy the `AssetServer` requirements.
+/// A dummy Asset reader, just to satisfy the `AssetServer` requirements.
 /// This implementation does nothing.
 pub(crate) struct DummyAssetIo {}
```
haihala, line 210: I'm pretty sure this could be deleted, but didn't do it yet.

src/spawn.rs
```diff
@@ -890,26 +910,24 @@ mod test {
...
                             &Visibility,
-                            &ComputedVisibility,
+                            &InheritedVisibility,
                             &ParticleEffect,
                             Option<&EffectSpawner>,
                         )>()
                         .iter(world)
                         .next()
                         .unwrap();
+                let visible = inherited_visibility.get();
```
haihala, line 925: Should probably check both `InheritedVisibility` and `ViewVisibility`, but couldn't figure out how to make tests work with this.

src/spawn.rs
```diff
@@ -561,24 +561,31 @@ pub fn tick_spawners(
...
             continue;
         };
 
         if asset.simulation_condition == SimulationCondition::WhenVisible {
-            if let Some(computed_visibility) = maybe_computed_visibility {
-                if !computed_visibility.is_visible() {
+            if let (Some(inherited_visibility), Some(_view_visibility)) =
+                (maybe_inherited_visibility, maybe_view_visibility)
+            {
+                // TODO: This should probably consider view visibility as well, but the tests don't
```
haihala, line 586: Should probably check both `InheritedVisibility` and `ViewVisibility`, but couldn't figure out how to make tests work with this.

src/render/mod.rs
```diff
@@ -2251,21 +2238,20 @@ pub(crate) fn queue_effects(
...
     trace!("queue_effects");
 
     // If an image has changed, the GpuImage has (probably) changed
     for event in &events.images {
         match event {
-            AssetEvent::Created { .. } => None,
-            AssetEvent::Modified { handle } => {
+            AssetEvent::Added { .. } | AssetEvent::LoadedWithDependencies { .. } => None,
+            AssetEvent::Modified { id } | AssetEvent::Removed { id } => {
+                // TODO: is fetching the handle necessary?
```
haihala, line 2250: I don't understand this well enough to make a call one way or another.

src/render/mod.rs
```diff
@@ -2551,7 +2505,10 @@ pub(crate) fn queue_effects(
                     pipeline: id,
                     entity,
                     sort_key: batch.z_sort_key_2d,
-                    batch_range: None,
+
+                    // FIXME: Guessing here
```
haihala, line 2509: No idea what these are supposed to be and the documentation seems lacking. Ditto for `Transparent3d` and `AlphaMask3d`.

src/spawn.rs
```diff
@@ -561,24 +561,31 @@ pub fn tick_spawners(
     mut query: Query<(
         Entity,
         &ParticleEffect,
-        Option<&ComputedVisibility>,
+        Option<&InheritedVisibility>,
+        Option<&ViewVisibility>,
```
haihala, line 565: Not sure why optional

djeedai: This is a discussion I've had a few times, Bevy tend to prefer inserting a component even if useless (user never intend to change visibility) whereas I consider that if you add a visibility component it's because you have an intent to toggle visibility, otherwise you shouldn't add the component in the first place (and the engine should assume everything visible by default). So I use `Option<>` in Hanabi for this. There's no consensus about that pattern as far as I know.

djeedai:
Thanks a lot @haihala for the PR. Given the timeline (Bevy 0.12 released yesterday, I just found out) and the complexity of the change (many new things in asset management and rendering, affecting severely the crate), I went ahead and I'm doing an upgrade on my side. I've looked at the PR, there are a few useful things, but some are incorrect I think (like using `AssetId` instead of `Handle`). I'll make a separate PR for the upgrade soon.

djeedai:
See #249

PR #254, opened 2023-11-19
Register `CompiledParticleEffect` type to inspect properties

ActuallyHappening:
Thanks so much for this awesome crate!

I was wanting to experiment with custom properties, which are awesome by the way, but realised that `CompiledParticleEffect` didn't implement `Reflect`.
This PR simply implements `Reflect` (not `FromReflect`) and adds the type into the registry, so that I can experiment with different values at runtime.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1131,25 +1131,39 @@ impl EffectShaderSource {
 /// ([`Visibility::Visible`]).
 ///
 /// [`set_property()`]: crate::CompiledParticleEffect::set_property
-#[derive(Debug, Clone, Component)]
+#[derive(Debug, Clone, Component, Reflect)]
+#[reflect(from_reflect = false)]
 pub struct CompiledParticleEffect {
     /// Weak handle to the underlying asset.
     asset: Handle<EffectAsset>,
+
     /// Cached simulation condition, to avoid having to query the asset each
     /// time we need it.
+    #[reflect(ignore)]
     simulation_condition: SimulationCondition,
-    /// Handle to the effect shader for his effect instance, if configured.
+
+    /// Handle to the effect shader for this effect instance, if configured.
+    #[reflect(ignore)]
     effect_shader: Option<EffectShader>,
+
     /// Instances of all exposed properties.
     properties: Vec<PropertyInstance>,
+
     /// Force field modifier values.
+    #[reflect(ignore)]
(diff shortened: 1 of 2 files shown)


djeedai:
I'm fairly conflicted about that change. I think the issue is that `CompiledParticleEffect` mixes both internal auto-generated fields that the user shouldn't touch (even with a debug tool) at runtime, while it also holds the runtime array of properties, which is designed to be accessed and modified at runtime.

I think the better fix is to split those two, and only reflect runtime the properties. That would have the benefit of skipping all effects with no property when manipulating them; that feels like a better fit for the ECS model.

ActuallyHappening:
The way I implemented `Reflect` for `CompiledParticleEffect` means only two fields are accessible through reflect: .asset and .properties. I think what you are saying is that both those fields are private, but by implementing reflect they effectively become public. (the .asset field I don't need at all, and will `#[reflect(ignore)]` in a later commit if desired)

I agree it is not the best 'fit for the ECS model' and Rust privacy/visibility rules. But it is also the simplest, already works, and avoids duplicating data (or adding a new type). Would you be satisfied if it was feature gated behind `debug-runtime-properties`?

djeedai:
Ok so I looked into it and I have a local change with properties split from `CompiledParticleEffect`, and I like that approach a lot. So after that I looked at your use case @ActuallyHappening and tried to edit at runtime some property via `bevy-inspector-egui`. And there are many issues:
- Changing the values of a vector is a total pain. The issue is that Hanabi's `VectorValue` type stores the vector components in compressed form (`[u32; 4]`; basically 16 bytes of memory), which really doesn't play nice with the default reflection implementation automatically derived. That should be fine for `ScalarValue` and `MatrixValue` though.
- Changing the vector element type doesn't work, it (weirdly) has no effect.
- Changing the number of components makes the renderer panic. Kind of expected since the type will not match the generated shader anymore, but still a footgun.

Although I guess all of those are the usual "great powers great responsibilities" of using reflection.

djeedai:
@ActuallyHappening can you try #256 please, let me know if that works for you? If so I'd prefer to merge that change rather than your current one. Thanks! :)

ActuallyHappening:
Yes this works very well!
One minor detail which is quite inconsequential, please order the struct field definitions so that .def is below .value, so that the first number you see (and are likely to edit) is not the default value, but the actual value.
image

PR #256, opened 2023-11-25
Split runtime properties into `EffectProperties`

djeedai:
Split the dynamic runtime storage of property values away from `CompiledParticleEffect`, which is an implementation detail, into a new `EffectProperties` component. This component manages access to property values by the user, as well as automated synchronization with the underlying `EffectAsset` when the list of declared properties changes.

This change adds a new `EffectSystems::UpdatePropertiesFromAsset` set during which the automated property update is performed. The set runs in Bevy's own `UpdateAssets` schedule, before `Update`.

As a consequence, change detection on properties now works as expected, and properties are only (re-)uploaded to GPU when the new `EffectProperties` component changed, instead of every frame as this was the case before this change. The component is also reflected, allowing reflection-based access and mutations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,19 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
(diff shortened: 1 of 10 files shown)


ActuallyHappening:
Yes this works very well!
One minor detail which is quite inconsequential, please order the struct field definitions so that .def is below .value, so that the first number you see (and are likely to edit) is not the default value, but the actual value.
image

djeedai:
> One minor detail which is quite inconsequential, please order the struct field definitions so that .def is below .value, so that the first number you see (and are likely to edit) is not the default value, but the actual value.

I appreciate the layout of `bevy-inspector-egui` is pretty confusing, but that's an edge case (either you're using it for debugging, or you should probably write a more specific tool that doesn't allow editing the def) so I prefer to have the field in "semantic order" in code, for clarity: first the definition of the property, and then its runtime value.

PR #260, opened 2023-11-28
Add `OrientModifier::rotation`

### LINKED ISSUE #258: Sprite rotation on oriented effects
When billboarding effects, or using other `OrientModifier`s, there should be an easy way to rotate the quads.

Having an easy way to rotate effects can break up the repeating pattern of particle textures. Having a property for initial spawn rotation could also allow for some other effects like making particles spin

It's theoretically possible to manually set it after applying the `OrientModifier`, but this is far to complex to be useful in practice.

**Additional context**
In godot it's possible to set initial particle rotation, along with a curve for how it changes over time:
image

djeedai:
Add an optional in-plane rotation to the `OrientModifier` to allow rotating the particles within their oriented plane.

Add 4 new scalar float attributes `F32_0` to `F32_3` which have no specific meaning, and can be used for anything. Use one in the `billboard.rs` example to store a per-particle initial rotation, which is later read back and used to set `OrientModifier::rotation` and give some visual variations to the effect.

Change `Attribute::ALL` to private; replace with `Attribute::all()`. This makes it a non-breaking change to add more attributes or reorder them in the future.

Fixes #258

### DIFF
--- a/src/attributes.rs
+++ b/src/attributes.rs
@@ -1517,8 +1517,8 @@ mod tests {
 
     #[test]
     fn attr_from_name() {
-        for attr in Attribute::ALL {
-            assert_eq!(Attribute::from_name(attr.name()), Some(attr));
+        for attr in Attribute::all() {
+            assert_eq!(Attribute::from_name(attr.name()), Some(*attr));
         }
     }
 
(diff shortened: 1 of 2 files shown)


djeedai:
@NiseVoid this should work, if you want to try it before I merge?

NiseVoid:
Seems to work as I'd expect, the `F32_0` thing got me confused for a bit however, and maybe the documentation should mention it's in radians (at least I assume that's the case since that's how the example treats it)?

djeedai:
`F32_0` itself is just a value. The example indeed uses it as an angle in radians. Good call thought, the rotation parameter should have a comment about this.

PR #266, opened 2024-01-13
Removed unused type param on `EffectProperties::with_properties`

Seldom-SE:


djeedai (approved):
Woah, I'm surprised Clippy or even the compiler don't complain about that one. Good catch, thanks!

PR #267, opened 2024-01-14
Add a global time scale for all effects

### LINKED ISSUE #239: Ability to pause / unpause effects independently from Bevy Time
I would like to stop the particles in place on demand, when I want. I don't mean spawing / despawning them, I mean just stop them in place wherever they are, and later continue from where they were. 
I am able to achieve this by pausing / unpausing `Res<Time>` provided by Bevy itself, but the problem is that I can't move the camera anymore (and anything else that depends on Bevy's `Time` has to be paused too). So I would like the ability to just pause / unpause effects independently from pausing / unpausing Bevy's `Time`.

In my builder game when the player wants to add a new building, the game enters a new state where characters movement is paused, particle execution is paused etc. They then place the building where they want and go back to the `Play` state. With the current situation I can either pause the particles execution while the new building is being placed, but also block the camera movement (because it depends on Bevy's `time.delta_seconds()` as well), or pause neither, so the camera can be moved, but the particle effects continue to execute, which looks odd.

Perhaps there is already a way to achieve this and I have missed it. Alternatively it can stay as is, but I feel it's a bit limiting.

**Additional context**
I can provide a GIF, if what I described is not clear.

(by the way thanks a lot for the library, it's amazing and I recommend it whenever relevant)

jannik4:
~~Add a global time scale for all effects. Currently very WIP. Probably needs:~~
- ~~Some tests/examples~~
- ~~More docs~~

Now reimplemented (and documented) using the `Time<T>` resource from bevy.

(Could still need some examples/tests when we have settled on the design)

Closes #239. Pausing should be possible by setting `EffectTimeScale` to zero.

### DIFF
--- a/src/render/mod.rs
+++ b/src/render/mod.rs
@@ -48,7 +48,7 @@ use crate::{
     render::batch::{BatchInput, BatchState, Batcher, EffectBatch},
     spawn::EffectSpawner,
     CompiledParticleEffect, EffectProperties, EffectShader, EffectSimulation, HanabiPlugin,
-    ParticleLayout, PropertyLayout, RemovedEffectsEvent, SimulationCondition, SimulationSpace,
+    ParticleLayout, PropertyLayout, RemovedEffectsEvent, SimulationCondition,
 };
 
 mod aligned_buffer_vec;
--- a/src/render/mod.rs
+++ b/src/render/mod.rs
@@ -47,7 +47,7 @@ use crate::{
     next_multiple_of,
     render::batch::{BatchInput, BatchState, Batcher, EffectBatch},
     spawn::EffectSpawner,
-    CompiledParticleEffect, EffectProperties, EffectShader, EffectTimeScale, HanabiPlugin,
+    CompiledParticleEffect, EffectProperties, EffectShader, EffectSimulation, HanabiPlugin,
     ParticleLayout, PropertyLayout, RemovedEffectsEvent, SimulationCondition, SimulationSpace,
 };
 
@@ -117,14 +117,22 @@ pub enum EffectSystems {
 #[derive(Debug, Default, Clone, Copy, Resource)]
 pub(crate) struct SimParams {
     /// Current effect system simulation time since startup, in seconds.
+    /// This is based on the [`Time<EffectSimulation>`](EffectSimulation) clock.
     time: f64,
     /// Delta time, in seconds, since last effect system update.
     delta_time: f32,
 
-    /// Current effect system simulation time since startup, in seconds.
-    unscaled_time: f64,
-    /// Delta time, in seconds, since last effect system update.
-    unscaled_delta_time: f32,
+    /// Current virtual time since startup, in seconds.
+    /// This is based on the [`Time<Virtual>`](Virtual) clock.
+    virtual_time: f64,
+    /// Virtual delta time, in seconds, since last effect system update.
+    virtual_delta_time: f32,
+
+    /// Current real time since startup, in seconds.
+    /// This is based on the [`Time<Real>`](Real) clock.
+    real_time: f64,
+    /// Real delta time, in seconds, since last effect system update.
+    real_delta_time: f32,
 }
 
 /// GPU representation of [`SimParams`], as well as additional per-frame
@@ -138,12 +146,18 @@ struct GpuSimParams {
     ///
     /// This is a lower-precision variant of [`SimParams::time`].
     time: f32,
-    /// Delta time, in seconds, since last effect system update.
-    unscaled_delta_time: f32,
-    /// Current effect system simulation time since startup, in seconds.
+    /// Virtual delta time, in seconds, since last effect system update.
+    virtual_delta_time: f32,
+    /// Current virtual time since startup, in seconds.
     ///
     /// This is a lower-precision variant of [`SimParams::time`].
-    unscaled_time: f32,
+    virtual_time: f32,
+    /// Real delta time, in seconds, since last effect system update.
+    real_delta_time: f32,
+    /// Current real time since startup, in seconds.
+    ///
+    /// This is a lower-precision variant of [`SimParams::time`].
+    real_time: f32,
     /// Total number of effects to simulate this frame. Used by the indirect
     /// compute pipeline to cap the compute thread to the actual number of
     /// effect to process.
@@ -167,8 +181,10 @@ impl Default for GpuSimParams {
         Self {
             delta_time: 0.04,
             time: 0.0,
-            unscaled_delta_time: 0.04,
-            unscaled_time: 0.0,
+            virtual_delta_time: 0.04,
+            virtual_time: 0.0,
+            real_delta_time: 0.04,
+            real_time: 0.0,
             num_effects: 0,
             render_stride: 0,   // invalid
             dispatch_stride: 0, // invalid
@@ -181,8 +197,10 @@ impl From<SimParams> for GpuSimParams {
         Self {
             delta_time: src.delta_time,
             time: src.time as f32,
-            unscaled_delta_time: src.unscaled_delta_time,
-            unscaled_time: src.unscaled_time as f32,
+            virtual_delta_time: src.virtual_delta_time,
+            virtual_time: src.virtual_time as f32,
+            real_delta_time: src.real_delta_time,
+            real_time: src.real_time as f32,
             ..default()
         }
     }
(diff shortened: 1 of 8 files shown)


src/render/mod.rs
```diff
@@ -1327,9 +1343,10 @@ pub(crate) fn extract_effects(
     trace!("extract_effects");
 
     // Save simulation params into render world
-    let dt = time.delta_seconds();
-    sim_params.time = time.elapsed_seconds_f64();
-    sim_params.delta_time = dt;
+    sim_params.time += ***time_scale * time.delta_seconds_f64();
+    sim_params.delta_time = (***time_scale * time.delta_seconds_f64()) as f32;
```
djeedai, line 1347: Isn't this essentially re-implementing partially `Time<T>`? Could we instead simply use `Time<EffectSimulationTime>`? See my comment on #239 for details.

jannik4: Looks like yes, just had a proper look at `Time<T>` for the first time. Something like `Time<EffectSimulation>` seems like a good idea then.

I would be happy to try to implement this. I can then imagine a total of three "Time/Delta Time" values:

- `sim_params.real_time`: From `Time<Real>`
- `sim_params.virtual_time`: From `Time<Virtual>`
- `sim_params.time`: From `Time<EffectSimulation>`

Where `Time<EffectSimulation>` is then a scaled version of `Time<Virtual>`. This would allow pausing/scaling time by both bevy's `Time<Virtual>` and by hanabi's `Time<EffectSimulation>`. All built in shader code would then use `sim_params.time`, but users can also use the other two versions.

This should be quite flexible. (But also adds a bit complexity)

jannik4: Implementation wise this seems pretty straight forward:

- bevy provides [virtual_time_system](https://docs.rs/bevy/latest/bevy/time/fn.virtual_time_system.html)
- So we can then just update `Time<EffectSimulation>` after that by `Time<Virtual>::delta()`

djeedai: Sounds all good to me. `SimParams` is small and uploaded once per frame, not per effect, so I'm totally fine adding more stuffs to it (including real/virtual times). Chances are, with DMA copy alignment (256 bytes?) we're already copying zeros instead at the minute, so it will make no difference.

src/render/mod.rs
```diff
@@ -1298,7 +1331,9 @@ pub(crate) fn extract_effect_events(
 ///
 /// [`ParticleEffect`]: crate::ParticleEffect
 pub(crate) fn extract_effects(
-    time: Extract<Res<Time>>,
+    real_time: Extract<Res<Time<EffectSimulation>>>,
+    virtual_time: Extract<Res<Time<EffectSimulation>>>,
```
djeedai, line 1335: Wrong times here?

src/time.rs
```diff
@@ -0,0 +1,172 @@
...
+        Self {
+            paused: false,
+            relative_speed: 1.0,
+            effective_speed: 1.0,
+        }
+    }
+}
+
+/// All methods for the [`Time<EffectSimulation>`](EffectSimulation) clock.
+pub trait EffectSimulationTime {
```
djeedai, line 46: Why do we need this trait?

src/time.rs
```diff
@@ -0,0 +1,172 @@
...
+pub(crate) fn effect_simulation_time_system(
+    virt: Res<Time<Virtual>>,
+    mut effect_simulation: ResMut<Time<EffectSimulation>>,
+) {
+    let virt_delta = virt.delta();
+    let effective_speed = if effect_simulation.context().paused {
+        0.0
+    } else {
+        effect_simulation.context().relative_speed
+    };
```
djeedai, line 163: Maybe wrap into a `relative_speed()` function that checks `paused` internally, for clarity?

src/time.rs
```diff
@@ -0,0 +1,172 @@
...
+    let delta = if effective_speed != 1.0 {
+        virt_delta.mul_f64(effective_speed)
+    } else {
+        // avoid rounding when at normal speed
+        virt_delta
+    };
+    effect_simulation.context_mut().effective_speed = effective_speed;
```
djeedai, line 170: I don't understand, `effective_speed` was read directly from the simulation clock, and is re-written back into it, but doesn't seem to be affected in any way by the virtual clock. Is this missing some `* virtual_speed` somewhere?

src/spawn.rs
```diff
@@ -800,7 +800,7 @@ mod test {
         // app.add_plugins(DefaultPlugins);
         app.init_asset::<Mesh>();
         app.add_plugins(VisibilityPlugin);
-        app.init_resource::<Time>();
+        app.init_resource::<Time<EffectSimulation>>();
```
djeedai, line 803: I'd assume this *also* needs `init_resource::<Time>()` since we need the base `Virtual` time. This is an emty app for testing, so won't have `Time` by default.

djeedai (changes requested):
This looks good overall but there's a number of small issues I think. Maybe some test would help validate the behavior here? Thanks!

src/time.rs
```diff
@@ -0,0 +1,172 @@
...
+    fn unpause(&mut self);
+
+    /// Returns `true` if the clock is currently paused.
+    fn is_paused(&self) -> bool;
+
+    /// Returns `true` if the clock was paused at the start of this update.
+    fn was_paused(&self) -> bool;
+}
+
+impl EffectSimulationTime for Time<EffectSimulation> {
```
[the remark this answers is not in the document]
jannik4: Unfortunately not, because `Time<T>` is a foreign type

src/time.rs
jannik4: See above

src/render/mod.rs
jannik4: Fixed

src/spawn.rs
jannik4: `Time` is not need as the time is advanced manually in the tests. (It was the same before, I didn't change anything there)

src/time.rs
jannik4: `effective_speed` is a local variable a few lines above and is set depending on `paused` and `relative_speed`.

`effective_speed` is just zero when paused and `relative_speed` when running. It is and should be independent of the virtual time.

The only thing that is dependent on the virtual time is the resulting delta for that frame.

jannik4: This is pretty much the same as here (without the extra clamping):

https://github.com/bevyengine/bevy/blob/abb8c353f49a6fe9e039e82adbe1040488ad910a/crates/bevy_time/src/virt.rs#L229-254

src/time.rs
jannik4: There is already a `relative_speed()` function that always returns the relative speed regardless of `paused`. This is exactly what bevy does for `Time<Virtual>`. I think it makes sense to do it the same way.

src/time.rs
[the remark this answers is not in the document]
djeedai: I still don't understand how all that works. What the relation with your trait?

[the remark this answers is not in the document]
djeedai: Oh ok I think I got it. It's a bit too clever.

src/time.rs
djeedai: Yes but `Virtual` is a self-contained clock whereas this one is based on `Virtual`. What is strange to me is that we document "the final speed is the product of the two clock speeds" but then when we calculate we only store this clock speed in `effective_speed`, not the product. So it's "effective" because it takes into account pausing, but it's still not "effective" as in being the final value used in practice.

djeedai: You can't compare with Bevy code, there's no clock in Bevy which is relative to another clock as far as I know.

jannik4: Sure, `Time<Virtual>` is based on `Time<Real>`: https://github.com/bevyengine/bevy/blob/abb8c353f49a6fe9e039e82adbe1040488ad910a/crates/bevy_time/src/virt.rs#L271-279

jannik4: Most of the time code is literally copy-pasted from `Time<Virtual>` and adapted (e.g. remove clamping).

jannik4: Ah, I think I see your point: Because `Time<Virtual>` already has time scaling.

jannik4: This should fix it: [1f9a23e](https://github.com/djeedai/bevy_hanabi/pull/267/commits/1f9a23e81571a2a2afc002643eb30452c1d2c99a)

jannik4:
Rebased on main and added a test for the effect simulation time. Could you take another look at this, @djeedai?

src/spawn.rs
djeedai: Ah ok and we don't insert the `effect_simulation_time_system` system so we don't need to read `Time<Virtual>`.

src/time.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// Returns the speed the clock advances relative to the virtual clock, as [`f32`].
+    fn relative_speed(&self) -> f32;
+
+    /// Returns the speed the clock advances relative to the virtual clock, as [`f64`].
+    fn relative_speed_f64(&self) -> f64;
+
+    /// Returns the speed the clock advanced relative to your system clock in this update,
+    /// as [`f32`].
+    ///
+    /// Returns `0.0` if the either [`Time<Virtual>`](Virtual) or
```
djeedai, line 56:
```suggestion
    /// Returns `0.0` if either the [`Time<Virtual>`](Virtual) or the
```

djeedai (approved):
Looks all good, thanks for the update!

djeedai (approved):

PR #278, opened 2024-02-17
Split render orientation and size

### LINKED ISSUE #269: `SizeOverLifetimeModifier` disrupts `OrientModifier`
**Crate versions**
`bevy` version: 0.12.1
`bevy_hanabi` version: 0.9 and 0.10-0-dev

**Describe the bug**
If you have a `SizeOverLifetimeModifier` with `screen_space_size = true` then the `OrientModifier` in-plane rotation is ignored. If `screen_space_size = false` then the `OrientModifier` works as required.

**Expected behavior**
`SizeOverLifetimeModifier` and `OrientModifier` to be independent transforms.

**To Reproduce**
Patch `billboard.rs` like so:
```
diff --git a/examples/billboard.rs b/examples/billboard.rs
index 6d384b4..b6d5208 100644
--- a/examples/billboard.rs
+++ b/examples/billboard.rs
@@ -160,8 +160,8 @@ fn setup(
                 rotation: Some(rotation_attr),
             })
             .render(SizeOverLifetimeModifier {
-                gradient: Gradient::constant([0.2; 2].into()),
-                screen_space_size: false,
+                gradient: Gradient::constant([500.; 2].into()),
+                screen_space_size: true,
             }),
     );
```
The clouds no longer have a random rotation.

**Screenshots**
hanabi

djeedai:
Split the calculation of the particle orientation (generally, from `OrientModifier`) and its size (_e.g._ from `SizeOverLifetimeModifier`) in such a way they become independent, and we can mix and match orientations with screen space size or simulation space size.

Fix screen space size to avoid a perspective divide, which was effectively breaking the depth-independent property of the size.

Fixes #269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Moved most shared WGSL code into an import module `vfx_common.wgsl`. This requires using `naga_oil` for import resolution, which in turns means `naga` and `naga_oil` are now dependencies of `bevy_hanabi` itself.
 
+### Fixed
+
+- Fixed a bug where particle attributes used in the context of a function would emit invalid code. (#275)
+- Fixed a bug where the screen-space size of particles was not correctly computed, leading to small variations in size. The new code correctly sizes particles based on a screen pixel size. This may increase the actual size of particles, if a larger size had been previously used to compensate the error introduced by this bug. (#269)
+- Fixed a bug where screen-space size ignored the particle's local orientation. (#269)
+
 ## [0.9.0] 2023-12-26
 
 ### Added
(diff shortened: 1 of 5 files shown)


djeedai:
@brettwitty do you want to try this fix before I merge it? I think this works, I tried various combinations of screen space size ON/OFF, and local/global simulation space, as well as `OrientModifier` options. But it's possible I missed something from the simplified example of `billboard.rs`.

BrettWitty:
Yep I'll give it a go in about 10 hours from now (7pm AEST)

BrettWitty:
All looks good to me. Thanks!

djeedai:
Ok I'll fix the test and merge then, thanks a lot for your feedback!

BrettWitty:
Thanks for chasing it down, and thanks for the great library!

PR #283, opened 2024-02-24
Add a modifier to make particles round.

pcwalton:
This commit adds a `RoundModifier`, which allows particles to be made into [squircle]s without having to use a texture. Squircles are like rounded rectangles but are cheaper to evaluate. A configurable `roundness` parameter allows the roundness to vary smoothly from a sharp rectangle to a perfect ellipse.

Given x and y from (-1, 1), the equation of the shape of the particle is |x|ⁿ + |y|ⁿ = 1, where n = 2 / `roundness`.

The `firework` and `portal` examples have been updated to use the roundness modifier, as rounded particles generally look better than square ones in these scenarios.

[squircle]: https://en.wikipedia.org/wiki/Squircle


djeedai (approved):
That's interesting, I didn't expect this but it looks like a nice way to sidestep the issue of handling different outputs (mesh, decal, trail, _etc._). I guess for now this is fine since there's a single output type, although the fact it's a modifier will be a blocker since this setting is only available for quad particles (roundness doesn't make sense for a mesh for example). But we can remove that modifier later.

I just want to test locally see how it looks. Otherwise all good. Thanks!

djeedai (approved):

djeedai (approved):

PR #285, opened 2024-02-26
Add dynamic `vec2` and `vec3` constructors to the expression grammar.

pcwalton:
There's currently no way to combine multiple dynamic scalar expressions into a scalar. This causes many problems, including not being able to convert an angle to a 2D vector, as demonstrated in the new doctests. This commit adds a new binary operator, `vec2`, and a new ternary operator, `vec3`, for this. Both operators work on all scalar types.


djeedai (approved):

djeedai (approved):
Looks good, thanks for the contribution!

PR #294, opened 2024-03-02
Improve serialization format of `EffectAsset`

djeedai:
This change makes a few improvements to the serialized representation of an `EffectAsset`:
- `LiteralExpr`, `ExprHandle`, and `Module` are now `#[serde(transparent)]`, which makes them less verbose.
- `VectorValue` is serialized as its glam math type representation, making it a lot more readable in text format (_e.g._ RON).

Before:

```
module: (
    expressions: [
        Literal((
            value: Vector((
                vector_type: (
                    elem_type: Float,
                    count: 3,
                ),
                storage: (1067030938, 3227307213, 1118770935, 0),
            )),
        )),
        Literal((
            value: Vector((
                vector_type: (
                    elem_type: Bool,
                    count: 2,
                ),
                storage: (0, 4294967295, 0, 0),
            )),
        )),
        Binary(
            op: Add,
            left: (
                index: 2,
            ),
            right: (
                index: 1,
            ),
        ),
    ],
),
```

After:

```
module: [
    Literal(Vector(Vec3((1.2, -3.45, 87.54485)))),
    Literal(Vector(BVec2((false, true)))),
    Binary(
        op: Add,
        left: 2,
        right: 1,
    ),
]
```


djeedai:
@pcwalton for visibility. Curious how we can manage migration of any existing asset already serialized? Here it's relatively easy to fix by hand, but of course that's not scalable. I'd expect more changes like this coming up in the future to make the format easier to read and edit by hand (when using RON).

PR #297, opened 2024-03-11
Flush `Commands` in between `prepare_effects` and `queue_effects`.

pcwalton:
This was causing nondeterministic failures due to the commands that `prepare_effects` added not running before `queue_effects`. Using the `.after()` method explicitly ensures that they're applied.

### DIFF
--- a/src/plugin.rs
+++ b/src/plugin.rs
@@ -205,7 +205,9 @@ impl Plugin for HanabiPlugin {
                 Render,
                 (
                     prepare_effects.in_set(EffectSystems::PrepareEffectAssets),
-                    queue_effects.in_set(EffectSystems::QueueEffects),
+                    queue_effects
+                        .in_set(EffectSystems::QueueEffects)
+                        .after(prepare_effects),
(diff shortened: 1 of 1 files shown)


src/plugin.rs
```diff
@@ -205,7 +205,9 @@ impl Plugin for HanabiPlugin {
                 Render,
                 (
                     prepare_effects.in_set(EffectSystems::PrepareEffectAssets),
-                    queue_effects.in_set(EffectSystems::QueueEffects),
+                    queue_effects
+                        .in_set(EffectSystems::QueueEffects)
```
djeedai, line 209: Shouldn't the system sets be ordered rather than the systems in them?

pcwalton: Yeah, so this is subtle. AFAICT the problem isn't the ordering; the systems *do* execute in the right order. The issue is that there's no Commands flush in between them. Adding `after` doesn't affect the order, but it does implicitly add a Commands flush to the schedule (per the documentation), fixing the bug.

PR #299, opened 2024-03-13
Stop double-drawing each particle group.

pcwalton:
We were looping over all the particle groups in `emit_draw`, but we already call that function once per group. So we were actually drawing each group twice.


djeedai (approved):
Good catch!

PR #300, opened 2024-03-13
Add methods to access and modify spawner count/time

NiseVoid:
I needed the ability to spawn an effect with a modified spawner, but these fields weren't accessible. This PR makes it possible to access those values.

I'm not sure if methods for this is the best approach, or if it makes more sense to make the fields public. I added both a getter method and set/with variants, since that matches the existing methods for `starts_active`.

### DIFF
--- a/src/spawn.rs
+++ b/src/spawn.rs
@@ -331,6 +331,54 @@ impl Spawner {
         Self::new(count, 0.0.into(), period)
     }
 
+    /// Set the number of particles that are spawned each cycle.
+    pub fn with_count(mut self, count: CpuValue<f32>) -> Self {
+        self.num_particles = count;
+        self
+    }
+
+    /// Set the number of particles that are spawned each cycle.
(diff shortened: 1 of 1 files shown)


djeedai:
@NiseVoid have you actually tried that change? It looks like `Spawner` is used to create `EffectSpawner` once in `tick_spawners()`, but after that is completely ignored. And there's no mutable access to the `EffectSpawner::spawner` field. So how are you planning on using those new functions? I'm not against the change, I'm just confused. 🤔

NiseVoid:
Yea I've tested it, it works fine. I use it to create a modified copy of an EffectAsset's Spawner and add that to the ParticleEffect. So basically like setting a property for an instance of an effect, but for the spawner

djeedai:
@NiseVoid see also #304, can't you directly spawn the `EffectSpawner` with the overrides, before Hanabi adds the component itself?

djeedai (approved):

PR #318, opened 2024-04-09
fix typo in `Attribute`

atornity:
I'm assuming these are supposed to be `2`?


djeedai (approved):
Yes I didn't pay attention 🙄 thanks for catching that!

PR #328, opened 2024-05-19
Fix indirect dispatch count when groups are used

### LINKED ISSUE #321: Cannot see more than 64 copies of Ribbon example particle effect
**Crate versions**
`bevy` version: 0.13.2
`bevy_hanabi` version: [2da6d8d](https://github.com/djeedai/bevy_hanabi/commit/2da6d8db1a02e028b279b5910eb2f1a75854c67c)

**Describe the bug**

Spawning 100 copies of the particle effect bundle from the Ribbons example leads to only 64 being visible

**Expected behavior**

100 instances of the particle effect are visible

**To Reproduce**

https://github.com/jakkos-net/bevy_hanabi_64_issue

**Screenshots**

image

djeedai:
Fix the number of workrgoups dispatched in the indirect dispatch pass when groups are used. Only the first group was counted.

Fixes #321

### DIFF
--- a/src/render/mod.rs
+++ b/src/render/mod.rs
@@ -3374,7 +3374,7 @@ impl Node for VfxSimulateNode {
             .write_buffer(render_context.command_encoder());
 
         // Compute init pass
-        let mut num_batches = 0;
+        let mut total_group_count = 0;
         {
             let mut compute_pass =
                 render_context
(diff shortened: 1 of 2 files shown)


djeedai:
This is a great stress test, thanks @jakkos-net!

image

PR #329, opened 2024-05-19
Add alternate effect to instancing example

NiseVoid:
The instancing example is very useful to reproduce bugs involving multiple effect instances or spawning/despawning, some of those however involve multiple effects. Thus to make the example more useful we add an alternate effect that is spawned with shift + space. The alternate effect has red circles instead of blue squares, it spins instead of expanding outwards, and has radial acceleration to at least showcase some update behavior.


djeedai (changes requested):
@NiseVoid can you please fix the `build_examples.[bat|sh]` scripts too? Thanks!

djeedai (approved):

PR #330, opened 2024-05-22
Fix effects getting stuck after a deletion

djeedai:
Fix some effects not updating after one or more effects have been deleted. This was due to the indirect pass shader assuming a contiguous `ParticleGroup` array, but that array contains gaps where effects have been deallocated, so updating a number of rows equal to the number of active effect means updating some unused rows and skipping some used ones in case there's any gap.


djeedai:
@NiseVoid if you want to try that change, it should fix effects getting stuck after another effect was deleted.

Issue #227, opened 2023-09-17
Allow customizing the blend function


SludgePhD:
**Describe the solution you'd like**
`bevy_hanabi` currently uses standard non-premultiplied alpha blending to draw all particles. It would be nice if that blend function was configurable instead of being hard-coded.

**Describe _why_ you want that solution. Is this related to a problem?**
I'm rendering my scene onto a transparent canvas, which is then blended onto the window. For that to look right, transparent objects need to use premultiplied alpha blending.

**Describe alternatives you've considered**
Deal with the graphical artifacts. This is what I'm doing now.

**Additional context**
Here's a smoke effect with my current setup. Above the line, the particles get drawn onto the transparent canvas, below the line, they get drawn onto an opaque sprite. There's a clearly visible seam in the particles that shouldn't be there.
![screenshot_2023-09-17_17:28:30](https://github.com/djeedai/bevy_hanabi/assets/96552222/a6238efd-3b19-406a-99cf-abcd9c6581ed)

I see that hanabi already supports alpha masking, and bevy's PBR pipeline supports alpha masking via the material's `AlphaMode`. Perhaps that type could be reused here?

PR #337, opened 2024-05-29
Add support for setting blending mode

Soulghost:
This commit adds a `with_blending_mode`, which can be used to set the blending mode for the particle renderer. In addition to alpha blend mode, premultiply and additive mode has been added.

The additive blending mode is useful for glow effects, the picture below makes a comparison.
### Additive Blending
image

### Alpha Blending
image

### Sample Code
```rust
//! Example of additive blend mode.
//!
//! This example demonstrate how to change the blend mode for the particle renderer.

use bevy::{
    core_pipeline::tonemapping::Tonemapping,
    log::LogPlugin,
    prelude::*,
    render::{
        camera::Projection, render_resource::WgpuFeatures, settings::WgpuSettings, RenderPlugin,
    },
};
#[cfg(feature = "examples_world_inspector")]
use bevy_inspector_egui::quick::WorldInspectorPlugin;

use bevy_hanabi::prelude::*;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let mut wgpu_settings = WgpuSettings::default();
    wgpu_settings
        .features
        .set(WgpuFeatures::VERTEX_WRITABLE_STORAGE, true);

    let mut app = App::default();
    app.insert_resource(ClearColor(Color::DARK_GRAY))
        .add_plugins(
            DefaultPlugins
                .set(LogPlugin {
                    level: bevy::log::Level::WARN,
                    filter: "bevy_hanabi=warn,additive=trace".to_string(),
                    update_subscriber: None,
                })
                .set(RenderPlugin {
                    render_creation: wgpu_settings.into(),
                    synchronous_pipeline_compilation: false,
                })
                .set(WindowPlugin {
                    primary_window: Some(Window {
                        title: "🎆 Hanabi — additive blending".to_string(),
                        ..default()
                    }),
                    ..default()
                }),
        )
        .add_plugins(HanabiPlugin);

    #[cfg(feature = "examples_world_inspector")]
    app.add_plugins(WorldInspectorPlugin::default());

    app.add_systems(Startup, setup)
        .add_systems(Update, bevy::window::close_on_esc)
        .run();

    Ok(())
}

fn setup(
    asset_server: Res<AssetServer>,
    mut commands: Commands,
    mut effects: ResMut<Assets<EffectAsset>>,
) {
    let camera = Camera3dBundle {
        transform: Transform::from_xyz(0.0, 0.0, 3.0).looking_at(Vec3::ZERO, Vec3::Y),
        projection: Projection::Perspective(PerspectiveProjection {
            fov: 90.0,
            ..Default::default()
        }),
        tonemapping: Tonemapping::None,
        ..Default::default()
    };

    commands.spawn(camera);

    let texture_handle: Handle<Image> = asset_server.load("orange_circle.png");

    let writer = ExprWriter::new();

    let age = writer.lit(0.).expr();
    let init_age = SetAttributeModifier::new(Attribute::AGE, age);

    let lifetime = writer.lit(5.).expr();
    let init_lifetime = SetAttributeModifier::new(Attribute::LIFETIME, lifetime);

    let init_pos = SetPositionCircleModifier {
        center: writer.lit(0.).expr(),
        axis: writer.lit(Vec3::Z).expr(),
        radius: writer.lit(0.2).expr(),
        dimension: ShapeDimension::Volume,
    };

    let init_vel = SetVelocityCircleModifier {
        center: writer.lit(Vec3::ZERO).expr(),
        axis: writer.lit(Vec3::Z).expr(),
        speed: (writer.lit(0.2) + writer.lit(0.2) * writer.rand(ScalarType::Float)).expr(),
    };

    // Use the F32_0 attribute as a per-particle rotation value, initialized on
    // spawn and constant after. The rotation angle is in radians, here randomly
    // selected in [0:2*PI].
    let rotation = (writer.rand(ScalarType::Float) * writer.lit(std::f32::consts::TAU)).expr();
    let init_rotation = SetAttributeModifier::new(Attribute::F32_0, rotation);

    let size = Vec2::splat(0.8);
    let effect = effects.add(
        EffectAsset::new(vec![32768], Spawner::rate(5.0.into()), writer.finish())
            .with_name("additive")
            .with_alpha_mode(bevy_hanabi::AlphaMode::Add)
            .init(init_pos)
            .init(init_vel)
            .init(init_age)
            .init(init_lifetime)
            .init(init_rotation)
            .render(ParticleTextureModifier {
                texture: texture_handle,
                sample_mapping: ImageSampleMapping::Modulate,
            })
            .render(SetSizeModifier { size: size.into() }),
    );

    commands
        .spawn(ParticleEffectBundle::new(effect))
        .insert(Name::new("effect"));
}
```

Add this image to the `assets/orange_circle.png`

orange_circle


SludgePhD:
Neat, this would fix https://github.com/djeedai/bevy_hanabi/issues/227

src/asset.rs
```diff
@@ -162,6 +162,24 @@ pub enum AlphaMode {
...
+    Alpha,
+
+    /// Similar to AlphaMode::Blend, however assumes RGB channel values are premultiplied.
+    Premultiply,
+
+    /// This blend mode simply adds pixel values of one layer with the other.
+    /// In case of values above 1 (in the case of RGB), white is displayed.
+    /// This is useful for glow effects, like those you might use for fire or magic spells.
+    Additive,
+}
```
SludgePhD, line 181: Any reason this enum is separate from `AlphaMode`? `bevy_pbr::AlphaMode` has all modes in a single enum.

Soulghost: Because we need to use a enum with Hash support for the ParticleRenderPipelineKey struct, but it seems that the bevy_pbr::AlphaMode has no Hash implementation. Maybe we can use the bevy_pbr::AlphaMode, and use a wrapper type for the ParticleRenderPipelineKey, let me try.

Soulghost: Since we already have a separated enum AlphaMode in bevy hanabi, Perhaps the best solution for now is to expand bevy_hanabi::AlphaMode, let me try.

Soulghost: I have merged the alpha blend options into bevy_pbr::AlphaMode, please help review the changes again, thanks!

examples/additive.rs
```diff
@@ -0,0 +1,123 @@
...
+    // spawn and constant after. The rotation angle is in radians, here randomly
+    // selected in [0:2*PI].
+    let rotation = (writer.rand(ScalarType::Float) * writer.lit(std::f32::consts::TAU)).expr();
+    let init_rotation = SetAttributeModifier::new(Attribute::F32_0, rotation);
+
+    let size = Vec2::splat(0.8);
+    let effect = effects.add(
+        EffectAsset::new(vec![32768], Spawner::rate(5.0.into()), writer.finish())
+            .with_name("additive")
+            .with_alpha_mode(bevy_hanabi::AlphaMode::Blend)
```
SludgePhD, line 107:
```suggestion
            .with_alpha_mode(bevy_hanabi::AlphaMode::Add)
```

src/render/mod.rs
```diff
@@ -1214,6 +1227,31 @@ impl SpecializedRenderPipeline for ParticlesRenderPipeline {
...
+                    operation: BlendOperation::Add,
+                },
+            },
+            AlphaBlendMode::Multiply => BlendState {
+                color: BlendComponent {
+                    src_factor: BlendFactor::Dst,
+                    dst_factor: BlendFactor::OneMinusSrcAlpha,
+                    operation: BlendOperation::Add,
+                },
+                alpha: BlendComponent::OVER,
```
SludgePhD, line 1251: Why is this using `OVER`, but the additive blend mode uses `dest.a = src.a + dest.a`?

Soulghost: It seems that there are different kinds of formulas for these kind of blending methods, in bevy we can find the Multiply Mode uses OVER, and the Additive Mode ignores the source alpha:

## Bevy
    - Multiply: https://github.com/bevyengine/bevy/blob/main/crates/bevy_pbr/src/render/mesh.rs#L1607
    - Additive: premultiplied and Add share the same pipeline key, They're made distinct in the PBR shader, via `premultiply_alpha()` 
        - https://github.com/bevyengine/bevy/blob/main/crates/bevy_pbr/src/material.rs#L469
        - https://github.com/bevyengine/bevy/blob/main/crates/bevy_pbr/src/render/mesh.rs#L1593
        - https://github.com/bevyengine/bevy/blob/main/crates/bevy_pbr/src/render/pbr_functions.wgsl#L692

## Unity
But in Unity there are some difference, the additive blend mode uses add for the alpha channel, but the multiply ignores the source alpha:
    - https://github.com/needle-mirror/com.unity.render-pipelines.universal/blob/8de5d7f7c63a433f5134fc8ce941072a08371ddc/Editor/ShaderGraph/Targets/UniversalTarget.cs#L1121

I seem to have mixed them up. Aligning with Bevy should be a good idea, I will modify them, thank you.

SludgePhD (approved):

Soulghost:
Hi @djeedai , Please help me review this PR, thanks!

src/render/mod.rs
```diff
@@ -1001,6 +1011,8 @@ pub(crate) struct ParticleRenderPipelineKey {
     /// Key: USE_ALPHA_MASK
     /// The effect is rendered with alpha masking.
     use_alpha_mask: bool,
+    /// The effect needs Alpha blend.
+    alpha_blend_mode: AlphaBlendMode,
```
djeedai, line 1015: Why are we keeping those separate, and using a separate `AlphaBlendMode`? Can't we just store `AlphaMode` like everywhere else?

src/render/mod.rs
```diff
@@ -1214,6 +1227,31 @@ impl SpecializedRenderPipeline for ParticlesRenderPipeline {
...
+            },
+            AlphaBlendMode::Multiply => BlendState {
+                color: BlendComponent {
+                    src_factor: BlendFactor::Dst,
+                    dst_factor: BlendFactor::OneMinusSrcAlpha,
+                    operation: BlendOperation::Add,
+                },
+                alpha: BlendComponent::OVER,
+            },
+        };
```
djeedai, line 1253: Can't we let the user provide all of that, and just store `AlphaMode::Blend(BlendState)`?

src/render/mod.rs
Soulghost: OK, I have modified the code to directly use `bevy_hanabi::AlphaMode`, but we must add Eq and Hash to the `bevy_hanabi::AlphaMode` struct since the ParticleRenderPipelineKey requires them.

Soulghost: Also, I have removed the full separate example and its assets.

src/render/mod.rs
Soulghost: Storing the blend state in the Blend member of the AlphaMode enum seemd like a good choice, and I initially tried to implement it. However, this approach appeared to introduce a breaking change to the interface. Therefore, I opted to add some additional options in parallel, similar to how bevy_renderer::AlphaMode is handled.

Soulghost: Also, the standard particle material's blending mode options appear to be sufficient according to the interfaces of Unity URP and Bevy. For more flexible customization, using custom materials may be a better choice.

djeedai: > this approach appeared to introduce a breaking change 

Bevy 0.14 releases in 2 weeks and will be a breaking change anyway. In general there's little point avoiding breaking changes until Bevy itself becomes stable.

Soulghost: Hi djeedai, it seems that we cannot just store `AlphaMode::Blend(BlendState)` since the BlendState struct is from wgpu, and it has no Eq and Hash trait implementations. So if we let the user provide all of that, we must duplicate a BlendState struct.

So is it a good choice to do not expose the BlendState directly to the user via the AlphaMode struct? If the user needs more flexible customization, we could provide some more powerful features such as custom materials in the future.

djeedai (approved):

djeedai:
@Soulghost would you mind please merge latest `main` to get the clippy fixes for the CI breaks ("Check format" action)? Ignore the "run-examples" failure, this is an upstream bug in `wgpu` it can't be fixed for now (it's fixed in Bevy 0.14).

Soulghost:
Hi @djeedai , I have rebased my feature branch based on the latest main branch, please review my code again, thanks a lot.

djeedai (approved):

Soulghost:
Hi @djeedai , it appears that the failure of `run-examples` and `coverage/coveralls` are preventing the merge process.

djeedai:
@Soulghost don't mind those. There's no auto-merge, I make the call on whether to merge. In that case coverage drop is minimal so all fine, and the failing examples is a bug upstream we can't do anything about for now. So all good. I'll have a final look later and merge.

Soulghost:
> @Soulghost don't mind those. There's no auto-merge, I make the call on whether to merge. In that case coverage drop is minimal so all fine, and the failing examples is a bug upstream we can't do anything about for now. So all good. I'll have a final look later and merge.

Got it, thank you very much.

PR #362, opened 2024-08-08
Support sampling normal distribution in expressions

### LINKED ISSUE #353: feat: Add an expression for normal distribution

I'd like to see a normal distribution equivalent to the existing [`uniform`](https://docs.rs/bevy_hanabi/latest/bevy_hanabi/graph/expr/struct.WriterExpr.html#method.uniform) method.

If I could simply call a `normal` method (or some other name) on a `ExprWriter`, the code would be much easier to read than stacking uniform distributions or some other work around to this.

This is this solution posted by NiseVoid on the discord.
image

**Additional context**

This is the distribution I'm thinking of.
image

msvbg:
Closes #353. I decided to look at this issue to build some familiarity with the code base. It mostly just copies what was done for the uniform distribution, with some minor tweaks.

I tested it by replacing various usages of `uniform` with `normal` in the examples, and it looked right to my eyes but I'm not sure how to test it more rigorously. I considered making a specialized example but that felt like overkill. Let me know what would pass the quality bar for you in terms of testing.


djeedai (commented):
Thanks for the contribution! Some minor improvement if you have time, would save a few instructions per sampling. Cheers

msvbg:
I don't fully understand this review comment, did you leave something out?

djeedai:
Yes no surprise because GitHub didn't post my actual comment.

The sampling for vec2 (and others) is too much. Box-Muller can create 2 random normal vars from 2 uniform ones, so you don't need to sample 2 vec2 vars, only 1 is enough. That would save some instructions. Same for vec3 and vec4.

msvbg:
Hey, sorry for the slow turnaround on this. I pushed a fix just now

djeedai (approved):

djeedai:
@msvbg there's a linter error, but should be fine after that. Thanks!

djeedai (approved):

PR #364, opened 2024-08-11
Add justfile for building wasm examples

Jengamon:
Just writes the build_examples_wasm.bat as a justfile to get cross-platform shelling for free (I think)

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,0 +1,197 @@
+export RUSTFLAGS := "--cfg=web_sys_unstable_apis"
+
+_default:
+    @just --list
+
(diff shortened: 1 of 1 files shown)


Jengamon:
understandable, but I don't really get the security concerns tho, as it is almost 1-to-1 with having a `bat` file in the repo. 

(the detail would be that https://github.com/casey/just usually shells out to `sh` and thus would be arguably more useful across systems. under the current system, Mac and Linux get the short end of the stick, and have to run the commands manually from the Windows batch file to build the warm examples)

PR #377, opened 2024-09-20
Allow particles to be arbitrary meshes.

pcwalton:
Allow particles to be arbitrary meshes.

Currently, Hanabi requires that particles be 2D quads. This is
sufficient for a good deal of VFX, but in many cases 3D objects are
required: smoke puffs, bullet casings, etc. This commit fixes this
deficiency by allowing particles to take on arbitrary meshes. To set the
mesh of a particle, use the new `EffectAsset::mesh` builder method. By
default, the mesh is a 2D quad.

The implementation is straightforward. The previously-hard-wired quad
vertices have been replaced with a `Handle<Mesh>`. The patch uses the
existing `bevy_render` infrastructure to upload the mesh to the GPU and
retrieve the vertices. Perhaps the most significant change is the
generalization of rendering to allow for indexed drawing in addition to
non-indexed. Because indexed drawing has a different on-GPU format for
indirect draw commands from that of non-indirect draw commands, some
additional bookkeeping is required.

This patch also adds support for a few features useful for 3D rendering:

* A `size3` attribute has been added, to allow the size to be controlled
in 3D.

* The `SetSizeModifier` now takes a 3D size gradient instead of a 2D
one.

* Vertex normals are available to modifiers via the `normal` shader
variable, as long as they call the new
`RenderContext::set_needs_normal` method.

A new example, `puffs`, has been added to demonstrate the use of 3D
meshes. It depicts the Bevy test fox running with cartoony smoke puffs
emitted at a constant rate behind it. Each puff consists of multiple
spherical mesh particles offset with some random jitter. A custom
Lambertian lighting modifier is supplied with the example, in order to
make the smoke puffs not appear solid white. (This modifier dramatically
improves the look of this example, but it's very limited, so I didn't
upstream it to Hanabi proper. A proper PBR lighting modifier would be
useful, but would be a significant amount of work, so I chose to defer
that to a follow-up.)


src/render/mod.rs
```diff
@@ -8,22 +8,16 @@ use std::{iter, marker::PhantomData};
...
-#[cfg(feature = "3d")]
-use bevy::{
-    core_pipeline::{
-        core_3d::{AlphaMask3d, Transparent3d},
-        prepass::OpaqueNoLightmap3dBinKey,
-    },
-    render::render_phase::{BinnedPhaseItem, ViewBinnedRenderPhases},
-};
 use bevy::{
+    core_pipeline::core_3d::Opaque3d,
```
mmatvein, line 12: This will have to be moved under the `#[cfg(feature = "3d")]` section below to make the CI pass.

src/render/mod.rs
```diff
@@ -1218,14 +1231,23 @@ impl SpecializedRenderPipeline for ParticlesRenderPipeline {
...
-                // Use depth buffer with alpha-masked particles, not with transparent ones
-                depth_write_enabled: key.use_alpha_mask,
+                // Use depth buffer with alpha-masked or opaque particles, not
+                // with transparent ones
+                depth_write_enabled: matches!(
+                    key.alpha_mask,
+                    ParticleRenderAlphaMaskPipelineKey::AlphaMask
+                        | ParticleRenderAlphaMaskPipelineKey::Opaque
+                ),
                 // Bevy uses reverse-Z, so Greater really means closer
```
mmatvein, line 1251: This same thing will have to be done for the section starting at line 1261 to make the CI pass.

mmatvein (approved):
I'm not an expert in the field, but I skimmed through the code and it looks good to me. I left two notes on things needed to get the CI to pass. It would be great to get this merged!

src/render/mod.rs
```diff
@@ -3412,8 +3527,23 @@ fn draw<'w>(
 
     pass.set_render_pipeline(pipeline);
 
+    let Some(gpu_mesh): Option<&GpuMesh> = meshes.get(&effect_batches.mesh) else {
+        return;
+    };
+
     // Vertex buffer containing the particle model to draw. Generally a quad.
-    pass.set_vertex_buffer(0, effects_meta.vertices.buffer().unwrap().slice(..));
+    pass.set_vertex_buffer(0, gpu_mesh.vertex_buffer.slice(..));
```
djeedai, line 3535: Just for my understanding, I assume Bevy is already managing storing the mesh into a vertex buffer? Is there any batching involved? (I assume there is from the `slice()` semantic)

src/render/vfx_render.wgsl
```diff
@@ -89,6 +102,15 @@ fn transform_position_simulation_to_world(sim_position: vec3<f32>) -> vec4<f32>
 #endif
 }
 
+fn transform_normal_simulation_to_world(sim_normal: vec3<f32>) -> vec3<f32> {
+#ifdef LOCAL_SPACE_SIMULATION
+    let transform = unpack_compressed_transform_3x3(spawner.transform);
+    return transform * sim_normal;
```
djeedai, line 108: Shouldn't normals be transformed by the inverse transpose of the transform? https://stackoverflow.com/questions/13654401/why-transform-normals-with-the-transpose-of-the-inverse-of-the-modelview-matrix

djeedai (changes requested):
Mostly all good, just the thing on the normal transformation, and the one about the mutable reference to `Assets`. Aside from that looks all good. Thanks once again! :)

src/render/mod.rs
pcwalton: Bevy manages storing the mesh, yeah.

In 0.14 there isn't any batching involved, but in 0.15 there is. (That is, in 0.14 the buffer is guaranteed to contain only one mesh; in 0.15 Bevy might have stuffed other meshes into the same buffer.) So this code will have to be significantly modified once we upgrade to 0.15. Basically just copy the code from the `Opaque3d` implementation in upstream Bevy.

src/plugin.rs
```diff
@@ -260,7 +260,11 @@ impl Plugin for HanabiPlugin {
             assets.insert(&HANABI_COMMON_TEMPLATE_HANDLE, common_shader);
         }
 
-        let effects_meta = EffectsMeta::new(render_device.clone());
+        let effects_meta = {
+            let mut assets = app.world_mut().resource_mut::<Assets<Mesh>>();
+            EffectsMeta::new(render_device.clone(), &mut assets)
```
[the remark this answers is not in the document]
pcwalton: I'm not sure what you mean here. `EffectsMeta` doesn't keep a reference to the assets: it only adds `default_mesh` to the assets and then saves its handle (not the mesh itself).

src/render/vfx_render.wgsl
pcwalton: Good point, fixed.

src/render/mod.rs
djeedai: Ah ok I missed that part sorry.

src/plugin.rs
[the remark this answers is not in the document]
djeedai: Ah ok I missed that part sorry.

pcwalton:
@djeedai Updated.

src/render/mod.rs
```diff
@@ -3468,8 +3583,23 @@ fn draw<'w>(
...
+    match gpu_mesh.buffer_info {
+        GpuBufferInfo::Indexed {
+            ref buffer,
+            count: _,
+            index_format,
+        } => {
+            pass.set_index_buffer(buffer.slice(..), 0, index_format);
+        }
+        GpuBufferInfo::NonIndexed => {}
+    }
```
djeedai, line 3602: The index buffer is already set below.

src/render/mod.rs
```diff
@@ -4380,6 +4548,18 @@ where
     first_buffer.expect("No buffers allocated")
 }
 
+impl ParticleRenderAlphaMaskPipelineKey {
```
djeedai, line 4551: Maybe use `impl From<LayoutFlags> for ParticleRenderAlphaMaskPipelineKey` instead? This sounds more idiomatic in Rust, no?

Cargo.toml
```diff
@@ -64,6 +64,11 @@ version = "0.14"
 default-features = false
 features = [ "bevy_core_pipeline", "bevy_render", "bevy_asset", "x11" ]
 
+[dependencies.bevy_gltf]
```
djeedai, line 67: Isn't that a `[dev-dependencies]` rather? It's only used in the `puffs.rs` example, isn't it?

Cargo.toml
```diff
@@ -160,6 +165,18 @@ required-features = [ "bevy/bevy_winit", "bevy/bevy_pbr", "3d" ]
 name = "ordering"
 required-features = [ "bevy/bevy_winit", "bevy/bevy_pbr", "3d" ]
 
+[[example]]
+name = "puffs"
+required-features = [
+    "bevy/bevy_winit",
```
djeedai, line 171: This is missing `bevy/png` to load the fox texture.

djeedai (changes requested):
Tried locally, the new puffs example is really nice!

Added a few comments, mostly minor things, but I think the normal transform is still wrong. Also can you please update the `CHANGELOG` as well as the various `build_examples` scripts?

If you're busy, I might consider merging as is and fixing after the fact, just because this touches quite a few files, to prevent further conflicts with other changes. Just let me know. Thanks!

src/render/vfx_render.wgsl
```diff
@@ -144,17 +171,22 @@ fn vertex(
...
-    let vpos = vertex_position * vec3<f32>(size.x, size.y, 1.0);
-    let sim_position = position + axis_x * vpos.x + axis_y * vpos.y;
+    let vpos = vertex_position * size;
+    let sim_position = position + axis_x * vpos.x + axis_y * vpos.y + axis_z * vpos.z;
     out.position = transform_position_simulation_to_clip(sim_position);
 
     out.color = color;
 
+#ifdef NEEDS_NORMAL
+    let normal = mat3x3(axis_x, axis_y, axis_z) * vertex_normal;
```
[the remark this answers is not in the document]
pcwalton: OK, I added an `inverse_transpose_mat3` function. I don't love having to do this in the vertex shader, but I didn't want to bloat the particles just for that.

pcwalton:
@djeedai Updated.

djeedai (approved):

PR #379, opened 2024-10-15
Possible typo?

kraftwer1:
I'm just going through your example – thanks a lot, I really enjoy learning WGSL through bevy_hanabi! One little thing that confused me: the comment states that there is a radius of "2 units", however the code says it's 0.5. Is there something I didn't get or is this a simple typo?


djeedai (changes requested):
Can you please also fix the top level comment in `lib.rs`? It contains a copy of that example. (I know, it's pretty bad maintenance-wise, sorry) Thanks!

djeedai (approved):
Thanks!

djeedai (approved):

PR #380, opened 2024-10-15
change the ribbon example into a tracer example.

naasblod:
I think this could be more helpful for people wanting to use the ribbon example. In my mind "tracers" are the most used application for this sort of thing, If you feel it isn't just close it.

run with:
 `cargo run --example ribbon --features="bevy/bevy_winit bevy/bevy_pbr 3d bevy/bevy_gizmos"`

### DIFF
--- a/examples/ribbon.rs
+++ b/examples/ribbon.rs
@@ -112,7 +112,6 @@ fn setup(mut commands: Commands, mut effects: ResMut<Assets<EffectAsset>>) {
 }
 
 fn move_head(
-    mut gizmos: Gizmos,
     mut query: Query<&mut Transform, With<ParticleEffect>>,
     mut effect: Query<&mut EffectProperties>,
     timer: Res<Time>,
@@ -129,7 +128,6 @@ fn move_head(
         ) * SHAPE_SCALE;
 
         properties.set("head_pos", (pos).into());
-        gizmos.sphere(pos, Quat::IDENTITY, 1.0, YELLOW);
         transform.translation = pos;
     }
 }
--- a/examples/ribbon.rs
+++ b/examples/ribbon.rs
@@ -1,12 +1,5 @@
-<<<<<<< HEAD
 //! Uses with_ribbons to draw a "tracer" or a "trail" following an Entity.
-=======
-//! Uses the [RibbonModifier] to draw a "tracer" or a "trail" following an Entity.
-//! The trail effect is achieved by using the [CloneModifier] on the "head" particle in combination
-//! with the [RibbonModifier].
->>>>>>> 8508ca1 (clean up and better description)
 //! The movement of the head particle is achieved by linking the particle position to a CPU position using a [Property] in [move_head].
-//!
 
 use bevy::color::palettes::css::YELLOW;
 use bevy::math::vec4;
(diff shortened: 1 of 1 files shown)


examples/ribbon.rs
```diff
@@ -125,9 +124,26 @@ fn setup(mut commands: Commands, mut effects: ResMut<Assets<EffectAsset>>) {
...
     for mut transform in query.iter_mut() {
-        transform.translation = vec3(f32::cos(theta), f32::sin(theta), 0.0) * 5.0;
+        let time = timer.elapsed_seconds() * TIME_SCALE;
+        let pos = vec3(
+            (1.0 - K) * (time.clone().cos()) + (L * K) * (((1.0 - K) / K) * time.clone()).cos(),
+            (1.0 - K) * (time.clone().sin()) - (L * K) * (((1.0 - K) / K) * time.clone()).sin(),
+            0.0,
+        ) * SHAPE_SCALE;
+
+        //let pos = vec3(f32::cos(theta), f32::sin(theta), 0.0) * 5.0;
```
djeedai, line 144: Delete this?

djeedai (commented):
If I understand correctly you're proposing to replace the purely GPU driven motion with a property-based one, to show how a ribbon can follow a target as a kind of trail, is that it?

I don't have any objection, but if you feel that tracers are common (I don't have an opinion, probably that's correct) then I'd expand a tiny bit the top level documentation to mention that this example shows a "tracer" (so that if someone search that term they find that example) and explain how a property is used to link the particle position to some CPU position (as this is probably a common feature to want).

examples/ribbon.rs
naasblod: Thanks for the comments I think those are great additions!

> replace the purely GPU driven motion with a property-based one
Yeah that's exactly it.

If you want I could expand on the example and make it separate to the ribbon if you feel like both are useful.
Meanwhile I'll crack on with the changes proposed here.

naasblod:
https://github.com/user-attachments/assets/82454667-9264-444a-ae42-5427b666a5f8

naasblod:
Is there anything else you'd like me to do with this or should I close it? Maybe move to a separate example?

pcwalton:
This is fine by me. Note that the little gaps between ribbon segments are something I've been thinking about fixing with the singly linked list experiment (that needs #387 to land first).

djeedai:
Apology @naasblod I had to prioritize merging the large fix from @pcwalton for ribbons due to its size and criticality, which now makes this change conflict. Please let me know if you have time ideally to fix, otherwise I'll try to find time to do it myself. Thanks for your understanding!

naasblod:
No worries friend ❤️! I'll take care of it!

naasblod:
I think that does it. Could use a final review

djeedai:
@naasblod the example fails to build:

```
error[E0412]: cannot find type `Gizmos` in this scope
   --> examples\ribbon.rs:115:17
    |
115 |     mut gizmos: Gizmos,
    |                 ^^^^^^ not found in this scope
```

naasblod:
@djeedai should be fixed now.

djeedai (approved):

djeedai:
@naasblod for next time, if you need a dependency for examples and tests only, you can add it to the `[dev-dependencies]` section in `Cargo.toml`.

naasblod:
Ah! Thanks, that's useful to know!

PR #383, opened 2024-10-17
Invalidate some bind group on buffer reallocate

djeedai:
Invalidate the sim bind group when the particle buffer is re-allocated. Also always re-create that bind group for now, as we don't invalidate on binding offset/size changes. With that, revert the fix for #330, which shouldn't be needed because the buffer is indeed contiguous.


djeedai:
@pcwalton and @NiseVoid, any chance you folks can try this change with your respective issue? For @pcwalton, the one we're discussing with the group atomic reorder, which this change doesn't fix but just to check that it's not making anything worse; and @NiseVoid, with the original change you tried which yielded the original #330 fix which is reverted here? I think this removes the confusion @pcwalton pointed and which is carried with code comment, that the buffer is indeed contiguous, which was assumed to be false in the fix for #330 (but was probably because the bind group was reusing an old buffer instead of the newly-allocated one). Thanks a lot in advance!

djeedai:
Note: review the instancing example before merging... it seems there's some issue sometimes with particles getting stuck, although it's unclear if this change is responsible (need to bisect) or it's a separate bug / a limitation due to capacity reached maybe?

djeedai:
Closing as obsolete in favor of a new upcoming rebased version. Suspecting this fixes #412 though.

PR #419, opened 2025-01-28
Replace ParticleEffectBundle with #[require] for ParticleEffect

### LINKED ISSUE #418: Transition from using Bundle to new "require" system
Use the "require" system as Bundles are deprecated.

Follows modern bevy style better and fits better into bevy 0.15^ code.
ParticleEffect CAN have requires for all bundled components, making ParticleEffectBundle essentially obsolete.

N/A

**Additional context**
It's a quick fix that im gonna start working on as of writing (mostly for fun).

kim-fg:
Completely removes the ParticleEffectBundle struct and replaces it with #[require(...)] for all required components.
All tests succeed after changing the usage to spawning ParticleEffect directly.

fixes #418 

(I also attempted to update the inline docs, but cant verify tests since DocTests simply wont run on my machine..)


kim-fg:
> There's not a 1:1 mapping between the bundle ("components which are generally useful") and the require mechanism ("mandatory components").
I understand what you mean, I can see how I made these changes without fully thinking it through.
However I don't understand, for example, how `Transform` can be "generally useful" and at the same time "not used". 
Either way, if this is not a wanted merge then you may of course close the PR and the linked issue.

(My stance is still that suggesting the use of the `Bundle` system is a bit confusing, but I'm not too attached.)

djeedai:
`Transform` is "generally useful" because that's almost always what users manipulate to set an entity transform in Bevy. It's "not used" (or rather, "not required") because what Hanabi requires and reads is `GlobalTransform`, which has been computed from `Transform` by Bevy. But there's no reason why someone shouldn't be able to use only `GlobalTransform` with Hanabi if they wish to do so. Putting `Transform` in the list of required component is over constraining; Hanabi never uses `Transform` directly.

djeedai:
Closing this, thanks for the PR! I've merged a similar change after reviewing the actual list of required components and cleaning it up. There was also a couple of traps in there, like the fact `Visibility` is mandatory due to some Bevy internals, and same for `Transform` unfortunately (Bevy doesn't work well if `GlobalTransform` is alone, the abstraction has some limitations).

PR #424, opened 2025-03-02
Hierarchical effects and GPU spawn event

djeedai:
This change introduces hierarchical effects, the ability of an effect to be parented to another effect through the `EffectParent` component. Child effects can inherit attributes from their parent when spawned during the init pass, but are otherwise independent effects. They replace the old group system, which is entirely removed. The parent effect can emit GPU spawn events, which are consumed by the child effect to spawn particles instead of the traditional CPU spawn count. Those GPU spawn events currently are just the ID of the parent particles, to allow read-only access to its attribute in _e.g._ the new `InheritAttributeModifier`.

The ribbon/trail system is also reworked. The atomic linked list based on `Attribute::PREV` and `Attribute::NEXT` is abandoned, and replaced with an explicit sort compute pass which orders particles by `Attribute::RIBBON_ID` first, and `Attribute::AGE` next. The ribbon ID is any `u32` value unique to each ribbon/trail. Sorting particles by age inside a given ribbon/trail allows avoiding the edge case where a particle in the middle of a trail dies, leaving a gap in the list.

A migration guide is provided from v0.14 to the upcoming v0.15 which will include this change, due to the large change of behavior and APIs.


CHANGELOG.md
```diff
@@ -19,16 +45,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   which is `None` if the layout is empty (as opposed to an empty string previously).
 - Removed effects are now deallocated from the render world before the extract schedule, instead of during it.
   This should have no consequence, unless you were using a system inserted explicitly before Hanabi's extraction.
+- `EffectAsset::capacities` is reverted to a singel `capacity: u32` value per effect.
```
tachimarten, line 48: typo: "single"

pcwalton, line 48: typo: "single"

PR #442, opened 2025-03-27
Add an `OnSpawn` condition for particle emission.

pcwalton:
[Unity has this.]

An example use case is spawning a bunch of particles as clumps around a circle. You make a hidden particle system A that uses the `SetPositionCircleModifier` to spawn invisible particles around a circle, then add an `EmitSpawnEventModifier` with an `OnSpawn` condition that triggers another particle system B to spawn, say, 5 particles for each particle in system A. Particle system B uses a `SetPositionSphereModifier` and inherits the position from particle system A. That way you get some nice clumps of particles all spaced randomly around a circle.

[Unity has this.]: https://docs.unity3d.com/6000.0/Documentation/ScriptReference/ParticleSystemSubEmitterType.Birth.html

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1146,7 +1146,7 @@ fn append_spawn_events_{0}(particle_index: u32, count: u32) {{
         if has_age {
             if has_lifetime {
                 age_code += &format!(
-                    "\n    let was_alive = particle.{0} < particle.{1};",
+                    "\n    let was_alive = particle.{0} < particle.{1};\n    let is_new = particle.{0} == 0.0;",
                     Attribute::AGE.name(),
                     Attribute::LIFETIME.name()
                 );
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1146,7 +1146,7 @@ fn append_spawn_events_{0}(particle_index: u32, count: u32) {{
         if has_age {
             if has_lifetime {
                 age_code += &format!(
-                    "\n    let was_alive = particle.{0} < particle.{1};",
+                    "\n    let was_alive = particle.{0} < particle.{1};\n    let is_new = particle.{0} == 0.0;",
```
djeedai, line 1149: This is wrong, there's nothing preventing the user from initializing `AGE` to something other than zero.

djeedai (commented):
I'm not convinced on this for 2 reasons:

1. Unity may have this, but the motivating example I think is wrong. Using a second invisible effect just to avoid some math in the spawn expression is an awful practice, which wastes so much GPU resources.
2. The proper and only way I think to detect a particle was spawned this frame is from the init pass, and so if we want to add this condition we should allow emitting GPU spawn events from the init shader of a parent. But that complexifies everything because parents are initialized after children, and GPU events are only remembered for a single frame.

pcwalton:
Closed in favor of #444, which subsumes this functionality.

PR #445, opened 2025-04-11
Fix visibility computation, allowing particles to appear in Bevy 0.16.

pcwalton:
There were two problems here:

1. The `ExtractComponentPlugin` for `EffectVisibilityClass` wasn't being
   added, so the render world never got a chance to see
   `EffectVisibilityClass`.

2. The Hanabi code was trying to query `RenderVisibleEntities` for
   `CompiledParticleEffect` components, but those APIs expect to take a
   visibility class instead.

I solved (1) by adding the plugin in the Hanabi plugin initialization,
and I solved (2) by switching `CompiledParticleEffect` calls to
`EffectVisibilityClass`. Making those changes causes particles to render
in Bevy 0.16.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,11 +100,11 @@ all-features = true
 
 [dev-dependencies]
 # For world inspector; required if "examples_world_inspector" is used.
-bevy-inspector-egui = "0.28"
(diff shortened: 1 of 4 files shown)


djeedai (approved):
Thanks for the fix!

PR #447, opened 2025-04-13
Recreate the init fill dispatch bind group when the arguments buffer resizes.

### LINKED ISSUE #446: Invalidate ChildInfo bind groups
**Crate versions**
main

**Describe the bug**
When you spawn a particle system, and then spawn the same particle system again, Hanabi crashes with `todo!("Invalidate ChildInfo bind groups")`. Commenting out that line doesn't fix the issue, it just makes it crash in wgpu instead ("Buffer with 'hanabi:buffer:child_infos' label has been destroyed").

This bug makes subemitters not really usable.

pcwalton:
This is currently not done, resulting in a crash when the arguments buffer resizes, which typically happens when multiple particle effects are spawned simultaneously.

There's a `todo!` that suggests that the original plan was to eagerly invalidate the init fill dispatch buffers when the arguments buffer changes size. I replaced this with usage of `CachedBindGroup` keyed off the ID of the arguments buffer, which seems to be the pattern used elsewhere in the code.

Closes #446.

### DIFF
--- a/src/render/event.rs
+++ b/src/render/event.rs
@@ -555,32 +555,8 @@ impl EventCache {
         self.init_indirect_dispatch_buffer
             .reserve(self.init_indirect_dispatch_buffer.len(), render_device);
 
-        let old_buffer = self.child_infos_buffer.buffer().map(|b| b.id());
         self.child_infos_buffer
             .write_buffer(render_device, render_queue);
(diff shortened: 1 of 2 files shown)


djeedai (approved):
Looks good I guess. It would be better if there was a way for me to test those changes though. Not sure if we can add an example or a test for them maybe?

PR #448, opened 2025-04-13
Take the source offset from the arguments into account in the init fill dispatch arguments shader.

pcwalton:
This was not done, frequently leading to particles not showing up if they weren't the first particle system spawned.

### DIFF
--- a/src/render/vfx_utils.wgsl
+++ b/src/render/vfx_utils.wgsl
@@ -82,7 +82,7 @@ fn init_fill_dispatch_args(@builtin(global_invocation_id) global_invocation_id: 
     // - bind the entire init indirect dispatch buffer (dst) and retrieve its offset here in the shader
     
     // Retrieve the source ChildInfo array buffer
-    let src = thread_index * args.src_stride;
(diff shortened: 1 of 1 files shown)


djeedai (commented):
I've looked into this in details. There's indeed a bug, and the fix probably work (I don't have a good way to test), but there's a lot more nonsense going on here. For example there's a bind group map from the event buffer index, but the init fill dispatch pass is copying from the child info buffer to the indirect dispatch one, so is completely independent of the event buffers themselves. Also the source and destination offsets already contain the proper values, so there's no need for an extra indirection inside the shader; this means that entire shader can disappear and we can use the normal `fill_dispatch_args`. Also the bind group uses 3 buffers but is only invalidated when some of them are re-allocated, but not others. Finally, we could batch everything together provided we ensure a 1:1 mapping between the index of the child info row and the index of the dispatch indirect row, which we currently don't but is not so complex to do. I'll try to finish my clean-up and merge it instead of this fix, because the current fix only solves one of the issues.

djeedai:
Fixed in #458.

PR #449, opened 2025-04-13
Use dynamic offsets for the arguments buffer when invoking the init fill dispatch shader.

pcwalton:
The existing code computes the dynamic offset for that uniform, but drops it on the floor without actually passing it to `set_bind_group`. This results in the init fill dispatch shader not actually being invoked on any particle system other than the first. This patch changes the code to pass the dynamic offset to the bind group, and also switches the init fill dispatch bind group layout to the one that takes dynamic offsets.


djeedai:
Fixed in #458.

PR #452, opened 2025-04-16
Add another pad variable.

pcwalton:
It's possible to run out if you use three `F32x3` attributes and velocity at the same time.


djeedai (approved):
We should find a way to make that dynamic, but good enough for now, thanks!

PR #455, opened 2025-04-19
Try reextracting next frame if an effect isn't loaded when Hanabi first encounters it.

pcwalton:
Hanabi uses the `Added<CompiledEffect>` query filter in order to determine when particle effects should be added to the `added_effects` table. But, if the particle effect asset isn't loaded at the time that query runs, Hanabi will skip adding the particle effect to that table. Because the `Added<CompiledEffect>` query filter will only match particle effects once, once the particle effect asset finally loads, it's too late to add it to the `added_effects` list. Thus it will never show up.

This patch fixes the issue by adding effects that are newly added but not yet fully loaded to a `pending_effects` list. The `added_effects` loop is modified to not only take into account effects that match the `Added<CompiledEffect>` filter but also effects in the `pending_effects` list. Effectively, that means if a particle effect isn't loaded yet, we poll it on subsequent frames until it finally is loaded before adding it to `added_effects`. This pattern is the same one used in many places in the Bevy renderer.

### DIFF
--- a/src/render/mod.rs
+++ b/src/render/mod.rs
@@ -1,4 +1,5 @@
 use std::marker::PhantomData;
+use std::mem;
 use std::{
     borrow::Cow,
     hash::{DefaultHasher, Hash, Hasher},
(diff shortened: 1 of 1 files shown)


djeedai (approved):
That's neat, I always wondered how Bevy was dealing with that issue (if at all). The `Local` pending queue makes a lot of sense. Thanks!

Issue #450, opened 2025-04-13
BufferId Assertion failed Panic when Instancing in multiple systems.


NicoZweifel:
**Crate versions**

`bevy` version:  0.15.3
`bevy_hanabi` version: 0.15.0

**Describe the bug**
`EffectAsset` gets created in a `Startup` system, referenced in a `Resource` and spawned in multiple systems (death, explosion etc), similar to the `Instancing` example:
```rs
#[derive(Resource)]
pub struct ExplosionParticleEffect {
    pub effect: Handle<EffectAsset>,
}
```
Then gets spawned like this in multiple systems:
```rs
        cmd.spawn((
            ParticleEffect::new(particle_effect.effect.clone()),
            Transform::from_translation(gtf.translation().with_y(1.0)),
            Lifetime {
                timer: Timer::from_seconds(2.0, TimerMode::Once),
            },
            Blood,
        ));
```
Effects get despawned recursively after a few seconds. It works fine for quite some time during gameplay until eventually, the application will panic:
```bash
thread '<unnamed>' panicked at C:\Users\nicoz\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\bevy_hanabi-0.15.0\src\render\mod.rs:5965:5:
assertion `left == right` failed
  left: BufferId(9)
 right: BufferId(21290)
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library\std\src\panicking.rs:695
   1: core::panicking::panic_fmt
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library\core\src\panicking.rs:75
   2: core::panicking::assert_failed_inner
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library\core\src\panicking.rs:420
... etc
```

**To Reproduce**
It used to work previously, the panics started happening after updating to `0.15.0`.

I have not managed to reproduce it in a stripped down example using the same effect and cleanup system. I haven't gotten it to panic even if I spawn multiple effects per frame in individual systems. Any Pointers/Ideas would be greatly appreciated. Thanks.

NicoZweifel:
It seems that this was related to a linking issue.

NicoZweifel:
[...]

PR #457, opened 2025-04-21
Remove potentially stale fields from `CachedEffect` & `EffectBatch`

NicoZweifel:
Resolves https://github.com/djeedai/bevy_hanabi/issues/450

### Notes/Improvements.

- `EffectBatch` could still contain `pub mesh_slice: Range<u32>`, which would result in one query per batch to `MeshAllocator` but the `indexed` information would have to be passed somewhere else, either pass `mesh_slice` and `indexed` to `from_input` on `impl EffectBatch` or pass only `mesh_slice` and save `indexed` as `bool` on the `CachedMesh`. I assume this probably isn't worth it atm since `MeshAllocator` access is immutable but I don't know for certain without testing/benchmarking.

### DIFF
--- a/src/render/batch.rs
+++ b/src/render/batch.rs
@@ -2,10 +2,7 @@ use std::{collections::VecDeque, fmt::Debug, num::NonZeroU32, ops::Range};
 
 use bevy::{
     prelude::*,
-    render::{
-        render_resource::{BufferId, CachedComputePipelineId},
-        sync_world::MainEntity,
-    },
+    render::{render_resource::CachedComputePipelineId, sync_world::MainEntity},
     utils::HashMap,
 };
 
@@ -90,12 +87,6 @@ pub(crate) struct EffectBatch {
     pub layout_flags: LayoutFlags,
     /// Asset ID of the effect mesh to draw.
     pub mesh: AssetId<Mesh>,
-    /// GPU buffer storing the [`mesh`] of the effect.
-    ///
-    /// [`mesh`]: Self::mesh
-    pub mesh_buffer_id: BufferId,
-    /// Slice inside the GPU buffer for the effect mesh.
-    pub mesh_slice: Range<u32>,
     /// Texture layout.
     pub texture_layout: TextureLayout,
     /// Textures.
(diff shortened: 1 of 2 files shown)


djeedai (commented):
@NicoZweifel this PR is now on the critical path as explained [on Discord](https://discord.com/channels/691052431525675048/1036006050526142514/1364355635822727249). Do you have time to finish it? Otherwise I'll take over, because this is an important bug/fix and there's a long list of things that depend on it (including a patch release, then a Bevy 0.16 release). Sorry about the pressure, no problem if you're not available, I just want to plan things. Thanks!

NicoZweifel:
Hey no worries, I've got some time and I can take a look at the discord. There was still one remaining issue (it doesn't crash but the particles are sort of distorted when it used to crash before). I was writing another comment about that but saw another PR got merged that might fix that. I can resolve the conflicts and go from there.

NicoZweifel:
@djeedai 
I've resolved the conflicts. Currently it doesn't panic but the linked [issue](https://github.com/djeedai/bevy_hanabi/issues/450) isn't fully resolved:

https://github.com/user-attachments/assets/439106cd-2a29-4b9c-9194-7cb66f07a1b3


At the moment I'm not sure if this is still the same issue just behaving differently because of my changes or if there could be stale property values as well, but I can invest some more time in the following days.

I took a look at `PropertyCache` and `HybridAlignedBufferVec`, could this be caused by the `range` beeing set in `insert` becoming stale?

```rs
let range = buffer.allocate(property_layout);
//...
CachedEffectProperties {
  buffer_index: buffer_index as u32,
  range,
}
```

djeedai:
From the video, my gut feeling is that it's not related to properties, because it seems the particles stop updating after spawn. Unless the particle position is driven by properties, I'd expect property issues to _e.g._ fail the spawn or spawn at the wrong location of with the wrong color, but not stop suddenly in the middle of the motion. But I may be wrong.

djeedai:
(that is not to say there's no property buffer bug; there may well be)

djeedai:
I have a fix in #461, which builds on top of this PR, and extend it with a proper invalidation based on comparing the mesh location (buffer ID and range, vertex and optionally index) cached vs. returned by the `MeshAllocator`. I ran your stress test @NicoZweifel and I could see the buffer reallocated (new `debug!` log), but no more visual artifacts, so I believe this solves it.

PR #458, opened 2025-04-21
Fix `InitFillDispatchArgs` source offset

djeedai:
Fix the source buffer offset for the init fill dispatch pass reading from the `GpuChildInfo` array the event count, and writing into the corresponding `GpuDispatchIndirect` struct the number of workgroups to dispatch to handle that number of events.

Remove the incorrect dependency to event buffers; the event buffers only store the events themselves, which are unused here. Instead, the init fill dispatch reads the event count, which is stored in the `GpuChildInfo` in a different buffer.

Remove the dedicated `init_fill_dispatch` compute kernel; instead we can use the generic `fill_dispatch` one since we're directly reading from and writing to uniform arrays.

Create a separate batching queue type `InitFillDispatchQueue` for the indirect init pass, which batches entries before they're submitted into the generic queue of buffer operations. Remove all sorting from that generic collection, now renamed `GpuBufferOperations`. Add support for separate queues of operations submitted as transactional units and dispatched as a single compute workload.

Fix `AlignedBufferVec::range_binding()` using the wrong (unaligned) item size, leading to an incorrect binding.

### DIFF
--- a/src/plugin.rs
+++ b/src/plugin.rs
@@ -28,12 +28,13 @@ use crate::{
         add_effects, batch_effects, clear_all_effects, extract_effect_events, extract_effects,
         fixup_parents, on_remove_cached_effect, on_remove_cached_properties, prepare_bind_groups,
         prepare_effects, prepare_gpu_resources, prepare_property_buffers, queue_effects,
-        resolve_parents, DebugSettings, DispatchIndirectPipeline, DrawEffects, EffectAssetEvents,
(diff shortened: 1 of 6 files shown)


djeedai:
@pcwalton this is a clean-up of the init fill dispatch logic, which should supersede both #448 and #449. It fixes a number of other small issues, notably around buffer re-allocations, as well as improves the chances of batching things together and reducing the number of compute dispatches. All examples run clean, but if you have a more complex test case to validate that'd be nice.

PR #468, opened 2025-05-06
Add `particle` access in the fragment shader

ntibi:
Allow `RenderModifier`s fragment shader to access the `Particle` buffer (just like the vertex shader) with `RenderContext::set_needs_particle_fragment`

My use case is sampling a small part of a noise texture for each rendered particle
I want to sample a different part of the noise texture according to a seed
So i need access to the particle custom attributes in the fragment shader

let me know if it's worth upstreaming

### DIFF
--- a/src/render/mod.rs
+++ b/src/render/mod.rs
@@ -1851,6 +1851,9 @@ pub(crate) struct ParticleRenderPipelineKey {
     /// Key: NEEDS_NORMAL
     /// The effect needs normals.
     needs_normal: bool,
+    /// Key: NEEDS_PARTICLE_IN_FRAGMENT
+    /// The effect needs access to the particle index and buffer in the fragment shader.
+    needs_particle_fragment: bool,
     /// Key: RIBBONS
     /// The effect has ribbons.
     ribbons: bool,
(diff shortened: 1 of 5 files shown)


src/render/mod.rs
```diff
@@ -1915,7 +1919,7 @@ impl SpecializedRenderPipeline for ParticlesRenderPipeline {
             // @group(1) @binding(0) var<storage, read> particle_buffer : ParticleBuffer;
             BindGroupLayoutEntry {
                 binding: 0,
-                visibility: ShaderStages::VERTEX,
+                visibility: ShaderStages::VERTEX_FRAGMENT,
```
ntibi, line 1922: it could be conditional (here and in the `EffectBuffer`) according to the pipeline key

djeedai: I don't have a good idea of the implications here, especially regarding performance, of making the bind group visible to the fragment shader or not.

djeedai:
Need to merge latest `main` to fix the linter isse.

src/render/mod.rs
ntibi: idk either about the theoretical cost of making a bind group visible

i did some frame sampling using renderdoc on the `instancing` example
to get a rough idea of the performance impact the binding visibility could have over the fragment shader

main (`visibility: VERTEX_FRAGMENT`):
108184 fragment calls in 87.39µs
=> 80 ns/fragment shader call

fork (`visibility: VERTEX_FRAGMENT`):
111944 fragment calls in 88.48µs
=> 79 ns/fragment shader call

it seems that having the bind group visible in the fragment shader has no impact on performance
i dont think it's necessary to make it conditional performance-wise

PR #480, opened 2025-06-10
Allow multiple effects to be packed into a single buffer again.

pcwalton:
This commit restores the functionality present in older versions of Hanabi that allows multiple effects to be packed into a single buffer. This is in preparation for reenabling batching.

I use the `base_instance` field on the effects metadata to supply the offset within each indirect buffer to the shader. This is convenient, because it naturally paves the way to use multi-draw as a follow-up, and it avoids having to add more fields to the effects metadata.

As suggested in the code, I rewrote sorting to be prior to batching instead of after batching. It now uses a modified version of [Tarjan's algorithm] to sort the effects into the optimum batching order. Not only does it make sure that parents precede children, but it also groups buffers together, and sorts by offset, in order to make batching easier as a follow-up. Sorting by offset within the buffer should enable the use of binary search in order for the shader to map from the global thread ID to the effect instance.

[Tarjan's algorithm]: https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search

### DIFF
--- a/src/render/batch.rs
+++ b/src/render/batch.rs
@@ -208,28 +208,36 @@ impl EffectSorter {
             entity_to_index.insert(effect.entity, index);
         }
 
-        // Next, create a map of parents to children.
-        let mut parent_to_children: Vec<_> = (0..self.effects.len()).map(|_| vec![]).collect();
+        // Next, create a map of children to their parents.
+        let mut children_to_parent: Vec<_> = (0..self.effects.len()).map(|_| vec![]).collect();
         for (kid, parent) in self.child_to_parent.iter() {
-            parent_to_children[entity_to_index[parent]].push(entity_to_index[kid]);
+            let (parent_index, kid_index) = (entity_to_index[parent], entity_to_index[kid]);
+            children_to_parent[kid_index].push(parent_index);
         }
 
-        // Now topologically sort the graph to determine the level of each node.
-        // This is a modification of the Tarjan algorithm that computes the
-        // depth of each node in the tree, not just the ordering.
+        // Now topologically sort the graph. Create an ordering that places
+        // children before parents.
         // https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search
-        let mut levels = vec![0; self.effects.len()];
+        let mut ordering = vec![0; self.effects.len()];
         let mut visiting = FixedBitSet::with_capacity(self.effects.len());
         let mut visited = FixedBitSet::with_capacity(self.effects.len());
         while let Some(effect_index) = visited.zeroes().next() {
             visit(
-                &mut levels,
+                &mut ordering,
                 &mut visiting,
                 &mut visited,
-                &parent_to_children,
+                &children_to_parent,
                 effect_index,
-                0,
             );
+        }
+
+        // Compute levels.
+        let mut levels = vec![0; self.effects.len()];
+        for effect_index in ordering.into_iter().rev() {
+            let level = levels[effect_index];
+            for &parent in &children_to_parent[effect_index] {
+                levels[parent] = levels[parent].max(level + 1);
+            }
         }
 
         // Now sort the result.
@@ -242,12 +250,11 @@ impl EffectSorter {
         // A helper function for topologically sorting the effect dependency
         // graph.
         fn visit(
-            levels: &mut Vec<u32>,
+            ordering: &mut Vec<usize>,
             visiting: &mut FixedBitSet,
             visited: &mut FixedBitSet,
-            parent_to_children: &[Vec<usize>],
+            children_to_parent: &[Vec<usize>],
             effect_index: usize,
-            current_level: u32,
         ) {
             if visited.contains(effect_index) {
                 return;
@@ -259,19 +266,12 @@ impl EffectSorter {
 
             visiting.insert(effect_index);
 
-            for &kid in &parent_to_children[effect_index] {
-                visit(
-                    levels,
-                    visiting,
-                    visited,
-                    parent_to_children,
-                    kid,
-                    current_level + 1,
-                );
+            for &parent in &children_to_parent[effect_index] {
+                visit(ordering, visiting, visited, children_to_parent, parent);
             }
 
             visited.insert(effect_index);
-            levels[effect_index] = current_level;
+            ordering.push(effect_index);
         }
     }
 }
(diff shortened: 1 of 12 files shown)


pcwalton:
@djeedai Thanks, it should be fixed now with the latest commit. This was because the sorting shaders previously weren't updated to deal with multiple effects in the same buffer. See the commit message for more details.

djeedai:
The ribbon example is completely broken.

https://github.com/user-attachments/assets/c6b835d2-c6d7-46ec-98de-d76d230e371c

pcwalton:
@djeedai Hmm, it works for me. What system are you on?

pcwalton:
`ribbon` is tested and is working on both Windows 11, NVIDIA GeForce RTX 4070 Laptop GPU and an macOS Apple M2 mac mini.

src/render/mod.rs
```diff
@@ -3933,9 +3965,8 @@ pub(crate) fn prepare_effects(
             handle: extracted_effect.handle,
             entity: extracted_effect.render_entity.id(),
             main_entity: extracted_effect.main_entity,
-            effect_slice,
+            effect_slice: effect_slice.clone(),
             init_and_update_pipeline_ids,
-            parent_buffer_index,
```
djeedai, line 3938: If we don't have the index of the parent buffer, we can't batch together updates for effect which have a parent, because the shader only has 1 bind point for a parent. It might not be used yet but conceptually we need that info for batching. Or am I missing something?

src/render/batch.rs
```diff
@@ -201,103 +148,131 @@ impl SortedEffectBatches {
...
 
-    /// Sort the effect batches.
-    pub fn sort(&mut self) {
-        self.sorted_indices.clear();
-        self.sorted_indices.reserve_exact(self.batches.len());
-
-        // Kahn’s algorithm for topological sorting.
+/// Sorts effects into the proper order for batching.
+///
+/// This places parents before children and also tries to place effects in the
```
djeedai, line 155: But this is not what we want, is it? The previous version sorts children before parents, because on frame N the update pass emits a GPU spawn event for a child effect, and on frame N+1 we need the child effect to get access to its parent particle buffer (to read _e.g._ the parent position) **before** the parent has a chance to init itself. Otherwise if the parent particle has died on frame N, we will recycle the entry in its particle buffer, and the child effect if processed after its parent will read a completely unrelated entry in the particle buffer.

src/render/batch.rs
```diff
@@ -429,91 +401,3 @@ pub(crate) struct InitAndUpdatePipelineIds {
     pub init: CachedComputePipelineId,
     pub update: CachedComputePipelineId,
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    fn make_batch(buffer_index: u32, parent_buffer_index: Option<u32>) -> EffectBatch {
```
djeedai, line 437: I'd rather keep tests like that. Best example why is that I tried to re-add it and that's how I realized:
1. the sorter in your change was sorting in the wrong order.
2. the algorithm is broken in the first place, doesn't even assign the correct levels to parents and children

src/render/batch.rs
```diff
@@ -201,103 +148,131 @@ impl SortedEffectBatches {
...
+        let mut visited = FixedBitSet::with_capacity(self.effects.len());
+        while let Some(effect_index) = visited.zeroes().next() {
+            visit(
+                &mut levels,
+                &mut visiting,
+                &mut visited,
+                &parent_to_children,
+                effect_index,
+                0,
+            );
```
djeedai, line 232: This algorithm simply doesn't work. I wrote a small test with 3 entities, inserted in that order:
- [0], child of [1]
- [1], parent of both [0] and [2]
- [2], child of [1]

Here's what happens:
- On first `while` iteration, index 0 is returned because it's never been visited. `visit()` then mark [0] as visited, skip the `for` loop since it has no children, and insert [0] with `level=0`.
- On second `while` iteration, index 1 is returned (the common parent). This time the `for` loop is executed, but child [0] is already visited so it's skipped. Child [2] is recursively visited with `level=1`. Finally parent [1] is inserted with `level=0`

We end up with:
- [0] level=0
- [1] level=0
- [2] level=1

This is wrong; [0] and [2] should have the same level, different (larger or smaller depending on convention) than their parent [1].

djeedai (changes requested):
See individual comments, there's multiple bugs in the sorter.

src/render/batch.rs
djeedai: <details>

<summary>Here's my test</summary>

```rust
#[cfg(test)]
mod tests {
    use super::*;

    fn insert_entry(
        sorter: &mut EffectSorter,
        entity: Entity,
        buffer_index: u32,
        base_instance: u32,
        parent: Option<Entity>,
    ) {
        sorter.effects.push(EffectToBeSorted {
            entity,
            base_instance,
            buffer_index,
        });
        if let Some(parent) = parent {
            sorter.child_to_parent.insert(entity, parent);
        }
    }

    #[test]
    fn toposort_batches() {
        let mut sorter = EffectSorter::new();

        // Some "parent" effect
        let e1 = Entity::from_raw(1);
        insert_entry(&mut sorter, e1, 42, 0, None);
        assert_eq!(sorter.effects.len(), 1);
        assert_eq!(sorter.effects[0].entity, e1);
        assert!(sorter.child_to_parent.is_empty());

        // Some "child" effect in a different buffer
        let e2 = Entity::from_raw(2);
        insert_entry(&mut sorter, e2, 5, 30, Some(e1));
        assert_eq!(sorter.effects.len(), 2);
        assert_eq!(sorter.effects[0].entity, e1);
        assert_eq!(sorter.effects[1].entity, e2);
        assert_eq!(sorter.child_to_parent.len(), 1);
        assert_eq!(sorter.child_to_parent[&e2], e1);

        sorter.sort();
        assert_eq!(sorter.effects.len(), 2);
        assert_eq!(sorter.effects[0].entity, e2); // child first
        assert_eq!(sorter.effects[1].entity, e1); // parent after
        assert_eq!(sorter.child_to_parent.len(), 1); // unchanged
        assert_eq!(sorter.child_to_parent[&e2], e1); // unchanged

        // Some "child" effect in the same buffer as its parent
        let e3 = Entity::from_raw(3);
        insert_entry(&mut sorter, e3, 5, 20, Some(e1));
        assert_eq!(sorter.effects.len(), 3);
        assert_eq!(sorter.effects[0].entity, e2); // from previous sort
        assert_eq!(sorter.effects[1].entity, e1); // from previous sort
        assert_eq!(sorter.effects[2].entity, e3); // simply appended
        assert_eq!(sorter.child_to_parent.len(), 2);
        assert_eq!(sorter.child_to_parent[&e2], e1);
        assert_eq!(sorter.child_to_parent[&e3], e1);

        sorter.sort();
        assert_eq!(sorter.effects.len(), 3);
        // Note: Ideally this would want e1 to be first and e3 after, so that e3 and e2
        // are grouped together since they're in the same buffer. But sorting by
        // topologicial order AND by buffer at the same time is in general impossible
        // (conflicting requirements; no solution).
        if sorter.effects[0].entity == e1 {
            assert_eq!(sorter.effects[1].entity, e3); // child first
        } else {
            assert_eq!(sorter.effects[0].entity, e3); // child first
            assert_eq!(sorter.effects[1].entity, e1); // child first
        }
        assert_eq!(sorter.effects[2].entity, e2); // parent after all children
        assert_eq!(sorter.child_to_parent.len(), 2);
        assert_eq!(sorter.child_to_parent[&e2], e1);
        assert_eq!(sorter.child_to_parent[&e3], e1);
    }
}
```

</details>

pcwalton: Are you sure that last test is right? We have:

    e1 -> e2 (i.e. e2 is a child of e1)
    e1 -> e3 (i.e. e3 is a child of e1)

    e1 buffer = 42
    e2 buffer = 5
    e3 buffer = 5

So a possible order is:

    e3, e2, e1

because this satisfies the constraints of children before parents (e3 and e2
before e1) and buffers grouped together (e2 = 5, e3 = 5). But this code doesn't admit that
order:

    // Note: Ideally this would want e1 to be first and e3 after, so that e3 and e2
    // are grouped together since they're in the same buffer. But sorting by
    // topological order AND by buffer at the same time is in general impossible
    // (conflicting requirements; no solution).
    if sorter.effects[0].entity == e1 {
        assert_eq!(sorter.effects[1].entity, e3); // child first
    } else {
        assert_eq!(sorter.effects[0].entity, e3); // child first
        assert_eq!(sorter.effects[1].entity, e1); // child first
    }

src/render/batch.rs
pcwalton: OK, it now sorts children before parents.

src/render/batch.rs
pcwalton: OK, this should be fixed now. Now it's two phases. First, I do a topological sort (in the reverse order I had before). Second, I assign levels, based on the constraint that, for each (parent, child) pair, the level of the parent must be at least the level of the child + 1. This seems correct.

src/render/batch.rs
pcwalton: I pushed a new commit that changes the test to match the comments. I *think* this is what you were going for.

pcwalton:
Marking this as a draft because I can reproduce the ribbon failure now.

pcwalton:
I've tested and `ribbon` is working on:
* AMD Ryzen 9 8945HS, NVIDIA GeForce RTX 4070 Laptop GPU, Vulkan
* AMD Ryzen 9 8945HS, NVIDIA GeForce RTX 4070 Laptop GPU, OpenGL
* 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz, NVIDIA GeForce RTX 2060, Vulkan
* 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz, NVIDIA GeForce RTX 2060, OpenGL
* Apple M2 Mac mini, Metal

Those are all the machines I have easy access to. It *doesn't* work on DX12, but that's expected because of the downlevel flag `VERTEX_AND_INSTANCE_INDEX_RESPECTS_RESPECTIVE_FIRST_VALUE_IN_INDIRECT_DRAW` (which I think is gone in newer `wgpu`).

djeedai:
Thanks for the extensive testing. I also couldn't reproduce after the first try, not sure why. But this looks like one of the bugs we had in the past, indexing being subtly wrong, but since it addresses a large buffer and doesn't go out of bounds, so we only see occasional visual artifacts. I will try to find some time this week or more likely week-end to debug once more, and if I don't find anything I'll merge as is.

djeedai:
@pcwalton I think we can close this PR, no? It should be entirely superseded by the existing batching code already in `main` I believe.

pcwalton:
Already implemented upstream.

PR #487, opened 2025-07-04
Put spawners into an array in the shader instead of using dynamic offsets.

pcwalton:
Currently, we use a dynamic offset to point to the `Spawner`. This prevents batching, so this PR changes the shaders so that the effect metadata contains an index to the spawner, and the shader picks the spawner out of an array.

This is built on top of https://github.com/djeedai/bevy_hanabi/pull/480, so only the last commit (61c205f) needs to be reviewed.


pcwalton:
I'm not sure why formatting is failing; it seems to be that perhaps the linter was upgraded?

djeedai:
Superseded by #525, closing.

PR #492, opened 2025-08-08
Add ColorByVelocityModifier

rydrman:
This modifier sets the color of a particle on a gradient based on its linear velocity at any point in time. Great for having particles disappear as they slow down, for example.

The implementation is largely a copy from the `ColorOverLifetimeModifier` but calculates the position in the gradient based on a clamped particle velocity

### DIFF
--- a/src/modifier/output.rs
+++ b/src/modifier/output.rs
@@ -366,6 +366,86 @@ impl RenderModifier for ColorOverLifetimeModifier {
     }
 }
 
+/// A modifier modulating each particle's color based on its velocity using a
+/// gradient curve.
+///
+/// # Attributes
+///
+/// This modifier requires the following particle attributes:
+/// - [`Attribute::VELOCITY`]
+#[derive(Debug, Default, Clone, PartialEq, Reflect, serde::Serialize, serde::Deserialize)]
+pub struct ColorByVelocityModifier {
+    /// The color gradient defining the particle color based on its lifetime.
+    pub gradient: Gradient<Vec4>,
+    /// The color blend mode.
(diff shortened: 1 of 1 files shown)


djeedai:
You can assign a color with a `SetAttributeModifier(Attribute::COLOR)` for example (or `HDR_COLOR` if you want `f32` color for HDR). It takes an expression which evaluates to the color value, so here a `u32` (LDR; RGBA8) or `vec4<f32>` (HDR). In your case you want that color to derive from velocity, which is read as `module.attr(Attribute::VELOCITY)` and returns a `Vec3`. From there you apply some `length()` expression to turn the velocity (vector) into speed (scalar), because I assume "color by velocity" really means "color by scalar speed" (it usually does). And to derive the actual color from the speed, apply any kind of gradient/LUT. I appreciate that last point is a bit lacking at the minute though. But I think I'd rather fix that gap, than add one more modifier to maintain. In the meantime, if you have a simple linear gradient you can build the conversion with expressions (mix the 2 color, using `clamp(speed, 0, max_speed) / max_speed` as the mix ratio in [0:1]).

djeedai:
Actually, you should be able to use a [`TextureSampleExpr`](https://docs.rs/bevy_hanabi/latest/bevy_hanabi/graph/expr/struct.TextureSampleExpr.html) to sample an image. Put the image in the `EffectMaterial` (add it if missing), don't forget to call `Module::add_texture_slot()`. I'd try with a "2D" texture/image (even if of size Nx1 pixels; but don't create it with Bevy as a 1D one) because I think Hanabi doesn't handle any other dimension yet.

PR #497, opened 2025-09-22
Check the results of `has_side_effect` for binary expressions.

pcwalton:
Currently, if the output of `UniformRand` is used in multiple dependent expressions, a new random number will be generated every time. This is at odds with the behavior of `BuiltInOperator::Rand`. The code *tries* to account for this by implementing `has_side_effect()` on binary expressions, but this is never actually checked. That's because `has_side_effect()` is only examined for *built-in* expressions, which `UniformRand` is not one of.

This commit factors the logic to create a local variable if there are no side effects out of `BuiltInExpr` into a separate function and calls that function in `Expr::Binary` and `BuiltInExpr`, fixing the issue.

A separate issue is that `Expr::has_side_effect()` is perhaps too aggressive, in that binary operators check to see whether their LHS and RHS are side-effecting and mark themselves as side-effecting if they are, when in fact only the operator itself determines whether the expression is side-effecting for the purposes of generating new temporaries. This causes many expressions to be marked as side-effecting that actually aren't. I opted not to fix this issue in this PR to keep this PR small, as it only results in more temporaries being generated, which shader compilers will clean up.

### DIFF
--- a/src/graph/expr.rs
+++ b/src/graph/expr.rs
@@ -1069,7 +1069,7 @@ impl Expr {
                 //     )));
                 // }
 
-                if op.is_functional() {
+                let body = if op.is_functional() {
                     if op.needs_type_suffix() {
                         let lhs_type = module.get(*left).and_then(|arg| arg.value_type());
                         let rhs_type = module.get(*right).and_then(|arg| arg.value_type());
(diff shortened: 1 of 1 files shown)


pcwalton:
The tests are failing because of the issue mentioned in the last paragraph in my commit message. Essentially, the tests never should have been passing, but this bug masked them. Fixing this bug made the tests start failing (in a way that's ultimately harmless to the end result) because additional temporaries were inserted. Let me know how you want me to proceed.

pcwalton:
@djeedai I can fix the test, but it'll be a little weird because the test will be checking that `a + b` yields something like `var1` instead of `var0 + var0`. Is that OK?

PR #498, opened 2025-09-22
Make sure `GpuEffectMetadata` gets initialized even if the mesh didn't change on the frame the effect was first extracted.

pcwalton:
It's possible for an effect to have a slot in the `GpuEffectMetadata` buffer reserved, but remaining uninitialized, if that effect has a `Visibility::Hidden` component on the first frame it was spawned. In such cases, Hanabi currently never initializes the `GpuEffectMetadata` at all, because it gates initialization of the metadata behind the mesh changing, which it won't in this case.

This commit fixes the bug by recording which effects have `GpuEffectMetadata` entries in a `MainEntityHashSet` in `EffectsMeta` and, if the effect doesn't have an entry in that set, creates a `GpuEffectMetadata` entry for it anyway, even if the mesh didn't change.

### DIFF
--- a/src/render/mod.rs
+++ b/src/render/mod.rs
@@ -14,6 +14,7 @@ use bevy::core_pipeline::core_2d::{Transparent2d, CORE_2D_DEPTH_FORMAT};
 use bevy::ecs::entity::EntityHashSet;
 #[cfg(feature = "2d")]
 use bevy::math::FloatOrd;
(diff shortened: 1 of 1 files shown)


djeedai (approved):
I think I noted that bug locally, but thanks for fixing! This is going to be replaced by the rewrite of extraction with more components and systems I'm working on, which makes it much easier to follow what's initializing what, but in the meantime the fix looks good. Thanks!

PR #507, opened 2025-10-17
Clarify OrientModifier per-instance rotation

barsoosayque:
More docs ! 

I've added an example and a note for `OrientModifier` to clarify how to use it to store per-instance rotation, since just using a regular expression such as `(writer.rand() * writer.lit(3.14)).expr()` is possible but wouldn't work as expected.

Example is basically a smol `examples/billboard.rs`.


djeedai (approved):
Looks good, thanks for clarifying!

PR #518, opened 2026-01-21
Bevy 0.18 update

morgenthum:
I updated bevy_hanabi in a one-shot prompt with gpt-5.2-codex because I need it for my game (and it works in my game). I reviewed the generated code, and it looks good to me. However, I am not deep into rendering, so I really can't tell if anything in render.rs is broken. Please be careful with this pull request! Just wanted to share it, if someone is interested.

All tests pass und the examples work.

### DIFF
--- a/src/plugin.rs
+++ b/src/plugin.rs
@@ -488,10 +488,10 @@ impl Plugin for HanabiPlugin {
                         // Upload and optionally resize the draw indirect args buffer
                         .after(update_mesh_locations)
                         // Bind groups depend on buffers being re-/allocated
-                        .before(EffectSystems::PrepareBindGroups),
+                        .before(prepare_bind_groups),
                     prepare_property_buffers
                         .in_set(EffectSystems::PrepareEffectGpuResources)
-                        .before(EffectSystems::PrepareBindGroups),
+                        .before(prepare_bind_groups),
                     prepare_effect_metadata
                         .in_set(EffectSystems::PrepareEffectGpuResources)
                         // Need DispatchBufferIndices to be allocated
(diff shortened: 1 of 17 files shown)


src/render/aligned_buffer_vec.rs
```diff
@@ -1196,7 +1196,7 @@ mod gpu_tests {
         queue.on_submitted_work_done(move || {
             tx.send(()).unwrap();
         });
-        let _ = device.poll(wgpu::PollType::Wait);
+        let _ = device.poll(wgpu::PollType::wait_indefinitely());
```
djeedai: This is a `wgpu` v28 construct, doesn't exist in v27.

src/plugin.rs
```diff
@@ -488,10 +488,10 @@ impl Plugin for HanabiPlugin {
                         // Upload and optionally resize the draw indirect args buffer
                         .after(update_mesh_locations)
                         // Bind groups depend on buffers being re-/allocated
-                        .before(prepare_bind_groups),
+                        .before(EffectSystems::PrepareBindGroups),
```
djeedai: Why did you change this?

djeedai (commented):
Thanks for the update, but I think the entire change is wrong.

As far as I understand Bevy has switched to managing internally `wgpu::BindGroupLayout` and encourages to use `PipelineCache` which takes a "key" (in the form of the descriptor) to lookup bind group layouts and lazy create them. However, Hanabi already has its own caching mechanism, and it looks like this change is basically adding Bevy's cache on top of Hanabi's, duplicating all the work.

Unfortunately I don't have time to go discuss with the rendering team trying to figure out how to make that change. If someone has context on how to approach that upgrade, I'm eager to hear about it.

src/render/aligned_buffer_vec.rs
morgenthum: You are absolutely right! As I said - this PR is AI generated, so we need to be careful. I fixed it.

src/plugin.rs
morgenthum: I fixed it.

morgenthum:
@pcwalton mentioned he also has a PR ready for the update. I think we should go with his version.

janhohenheim:
Tried this out for a bit and it seems to work for WebGPU! So for anyone wondering which Hanabi flavor to use, I recommend:
- native targets: https://github.com/pcwalton/bevy_hanabi
- WebGPU: this fork

djeedai:
Updated via #521, so closing this. Thanks for the initial work @morgenthum, this helped me with a few changes. In the end I had to switch to `BindGroupLayoutDescriptor` anyway, but there were a few more subtle issues to fix. And there were also some meta issues with _e.g._ CI and `wasm-bindgen` and `tarpaulin` versions.

Issue #510, opened 2025-10-25
Despawning effect hierarchy fail asserts


barsoosayque:
**Crate versions**
`bevy` version: _0.17_
`bevy_hanabi` version: _0.17.0_

**Describe the bug**
Despawning a root entity with the effect hierarchy (the one that targeted by `EffectParent`) results in numerous failed asserts, which in debug are just CTDs.

This one I can reproduce in `examples/firework.rs` (diff below):
```log
2025-10-25T08:39:19.418161Z ERROR bevy_hanabi::render: GPU fill dispatch buffer operation bind group not found for buffers src#BufferId(36) dst#BufferId(35)
2025-10-25T08:39:20.852071Z ERROR bevy_hanabi::render: GPU fill dispatch buffer operation bind group not found for buffers src#BufferId(36) dst#BufferId(35)

thread 'Compute Task Pool (9)' (76223) panicked at src/render/batch.rs:365:9:
assertion `left == right` failed
  left: false
 right: true
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `bevy_hanabi::render::batch_effects`!
Error: ExampleFailedError(1)
```

The second one I can only repro in my project's code, the same exact story but the assert is different:
```log
2025-10-25T08:47:04.567248Z TRACE bevy_hanabi::render: => ParticleEffect on render entity 3189v45 associated with main entity MainEntity(12690v1), removing...
2025-10-25T08:47:04.567272Z TRACE bevy_hanabi::render: => GPU particle slab #41 gone, destroying its bind groups...
2025-10-25T08:47:04.567360Z TRACE bevy_hanabi::render::event: Removing cached event CachedEffectEvents { buffer_index: 0, range: 1024..1280, init_indirect_dispatch_index: 4 } from cache.
2025-10-25T08:47:04.567383Z TRACE bevy_hanabi::render: => ParticleEffect on render entity 3190v33 associated with main entity MainEntity(12695v0), removing...
2025-10-25T08:47:04.567394Z TRACE bevy_hanabi::render: => GPU particle slab #37 gone, destroying its bind groups...
2025-10-25T08:47:04.567404Z TRACE bevy_hanabi::render::event: on_remove_cached_effect_events
2025-10-25T08:47:04.567423Z TRACE bevy_hanabi::render::event: Removing cached event CachedEffectEvents { buffer_index: 0, range: 1024..1280, init_indirect_dispatch_index: 4 } from cache.

[...]

PR #546, opened 2026-06-30
Fix crash when despawning parent effects

lkurcak:
Follow-up on #527 

Re-tested on the latest main and still seeing the modified fireworks example from #510 crash, whereas with my modification the crash no longer occurs.


djeedai:
Thanks @lkurcak for the PR and sorry I didn't reply earlier. That fix is a bit too broad I think, and destroys too many components. I have a more targeted fix in #561 which I believe is the minimal and correct fix for the GPU buffer double-free. I tested on `firework.rs` and that's the only bug I can repro on `main` (so, past 0.19). So I believe that fixes all of #510.

lkurcak:
Thanks for the response. Verified the targeted fix from #561 resolves my issue as well. Closing this as a result.

djeedai:
Thanks a lot for confirming!

PR #551, opened 2026-07-21
SetPositionBoxModifier

hukasu:
Create a `SetPositionBoxModifier`


Cargo.toml
```diff
@@ -67,7 +67,9 @@ wgpu = { version = "29.0.3", default-features = false, features = [
   "fragile-send-sync-non-atomic-wasm",
 ] }
 naga = { version = "29.0.3", features = ["wgsl-in"] }
-naga_oil = { version = "0.22", default-features = false, features = ["test_shader"] }
+naga_oil = { version = "0.22", default-features = false, features = [
```
djeedai: Avoid random formatting unrelated to the change please.

examples/box.rs
```diff
@@ -0,0 +1,134 @@
+//! An example using the [`SetPositionBoxModifier`].
```
djeedai: There's already the `init.rs` example showing all position modifiers; can you please update that example rather than adding a new one?

src/modifier/position.rs
```diff
@@ -343,3 +343,125 @@ impl Modifier for SetPositionCone3dModifier {
         Ok(())
     }
 }
+
+/// A modifier to set the position of particles on a box.
+///
+/// The box is axis aligned.
+///
+/// Particles are moved somewhere inside the volume or on the surface of a
+/// box define by its scale.
```
djeedai:
```suggestion
/// box defined by its scale.
```

src/modifier/position.rs
```diff
@@ -343,3 +343,125 @@ impl Modifier for SetPositionCone3dModifier {
...
+/// # Attributes
+///
+/// This modifier requires the following particle attributes:
+/// - [`Attribute::POSITION`]
+#[derive(Clone, Copy, Hash, Reflect)]
+pub struct SetPositionBoxModifier {
+    /// The scale of the box.
+    ///
+    /// Expression type: `Vec3`
+    pub scale: ExprHandle,
```
djeedai: Not sure "scale" is the best semantics here. I'd rather use something like "extents" or "half extents".

src/modifier/position.rs
```diff
@@ -343,3 +343,125 @@ impl Modifier for SetPositionCone3dModifier {
...
+    }}
+    (*particle).{} = vec3(x, y, z);
+"#,
+                            scale,
+                            Attribute::POSITION.name()
+                        )
+                    }
+                    ShapeDimension::Volume => format!(
+                        r#"    let scale = {};
+    let mult = vec3(frand(), frand(), frand()) - 0.5;
```
djeedai:
```suggestion
    let mult = frand3() - 0.5;
```

src/modifier/position.rs
```diff
@@ -343,3 +343,125 @@ impl Modifier for SetPositionCone3dModifier {
...
+///
+/// Particles are moved somewhere inside the volume or on the surface of a
+/// box define by its scale.
+///
+/// # Attributes
+///
+/// This modifier requires the following particle attributes:
+/// - [`Attribute::POSITION`]
+#[derive(Clone, Copy, Hash, Reflect)]
+pub struct SetPositionBoxModifier {
```
djeedai, line 359: I think we need a box center offset field. Most shape-based position and velocity modifiers have an offset. I know in general you can offset the emitter transform. But an implicitly centered box at (0,0,0) doesn't work if by any chance you want another implicitly-centered modifier with a different origin. So I think it's best to have an explicit one.

```suggestion
pub struct SetPositionBoxModifier {
    /// The box center, relative to the emitter position.
    ///
    /// Expression type: `Vec3`
    pub center: ExprHandle,
```

djeedai (changes requested):
Thanks for the PR! Approved conceptually, as this is one I was planning to do soon, however there's a number of changes needed with the actual code.

hukasu:
Screenshot for showcase 
image

i will be right back to finish the remaining suggestions

djeedai (approved):

PR #552, opened 2026-07-21
RotateOverTimeModifier

hukasu:
Very naive implementation of `RotateOverTimeModifier`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,9 @@ wgpu = { version = "29.0.3", default-features = false, features = [
   "fragile-send-sync-non-atomic-wasm",
 ] }
 naga = { version = "29.0.3", features = ["wgsl-in"] }
-naga_oil = { version = "0.22", default-features = false, features = ["test_shader"] }
+naga_oil = { version = "0.22", default-features = false, features = [
+  "test_shader",
+] }
 
 # getrandom requires both 'wasm_js' as a feature AND as a RUSTFLAGS to work on wasm
 [target.'cfg(target_arch = "wasm32")'.dependencies]
@@ -164,6 +166,9 @@ name = "puffs"
 
 [[example]]
 name = "lightning"
+
+[[example]]
+name = "rotate_over_time"
 
 [[test]]
 name = "empty_effect"
--- a/src/modifier/rotate_over_time.rs
+++ b/src/modifier/rotate_over_time.rs
@@ -1,0 +1,47 @@
+use bevy::reflect::Reflect;
+
+use crate::{
+    Attribute, BoxedModifier, BuiltInExpr, EvalContext, ExprError, ExprHandle, Modifier,
+    ModifierContext, Module, ShaderWriter,
+};
+
+/// Rotates particles over time.
+#[derive(Clone, Copy, Reflect)]
+pub struct RotateOverTimeModifier {
+    /// Rotation that the particle will have in a second.
+    ///
+    /// Expr type: Mat4
+    pub rotation: ExprHandle,
+}
+
+impl Modifier for RotateOverTimeModifier {
+    fn context(&self) -> ModifierContext {
+        ModifierContext::Update
+    }
+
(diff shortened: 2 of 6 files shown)


src/modifier/rotate_over_time.rs
```diff
@@ -0,0 +1,47 @@
...
+    Attribute, BoxedModifier, BuiltInExpr, EvalContext, ExprError, ExprHandle, Modifier,
+    ModifierContext, Module, ShaderWriter,
+};
+
+/// Rotates particles over time.
+#[derive(Clone, Copy, Reflect)]
+pub struct RotateOverTimeModifier {
+    /// Rotation that the particle will have in a second.
+    ///
+    /// Expr type: Mat4
```
djeedai: I don't think we want that `Mat4` field, users are already struggling with vector operations and complaining there's too many missing helpers, so for matrices where we have almost nothing, basically nobody will use this modifier. Not your fault, but basically I don't want to try encouraging users to use `Mat4` and drive them into a corner where there's no utilities to manipulate matrices; that will just lead to more opened issues and frustration.

src/modifier/rotate_over_time.rs
```diff
@@ -0,0 +1,47 @@
...
+        Box::new(*self)
+    }
+
+    fn apply(&self, module: &mut Module, context: &mut ShaderWriter) -> Result<(), ExprError> {
+        let rotation = context.eval(module, self.rotation)?;
+        let dt = BuiltInExpr::new(crate::graph::BuiltInOperator::DeltaTime).eval(context)?;
+        context.main_code += &format!(
+            r#"    {{
+        let rotation = {rotation};
+        particle.{0} = normalize(mix(particle.{0}, (vec4(particle.{0}, 1) * rotation).xyz, {dt}));
```
djeedai: For me `RotateOverTimeModifier` (and this is confirmed by its doc comment) means that the particle will continuously rotate. This is not the case; this modifier interpolates the rotation of the particle from its current rotation to a target one, then does nothing after that.

The other major issue is that calling `mix(a, b, dt)` only works nicely if `dt < 1`, otherwise the interpolation will overshoot. This is because we're mixing units; `dt` is in seconds here, whereas the `mix(.., .., r)` ratio is a unitless quantity in `0..=1`. If you really intend to create an interpolator modifier, it should take a `duration` parameter that determines how long it should take to go from current to target rotation, and then calculate the unitless `mix(.., .., r)` ratio such that it takes that much time relative to `dt`.

Cargo.toml
```diff
@@ -165,6 +167,9 @@ name = "puffs"
 [[example]]
 name = "lightning"
 
+[[example]]
+name = "rotate_over_time"
```
djeedai, line 171: In general please avoid adding examples unless there's a non-obvious new feature or there's really no other example this could fit in. The reason is that the maintenance cost of examples is quite high; not only are examples used as behavior benchmark to validate behavior for all major changes (need to run them all one by one for visual check), they also require extra work in `wasm` (add new webpage, extra compile time, _etc._). Ideally all of that should be automated or covered by unit/feature tests, but until then it's quite the burden. Thanks!

djeedai (commented):
I'm skeptical about this change. See individual comments. As is I'm not convinced.

src/modifier/rotate_over_time.rs
hukasu: The matrix indicates the rotation that the particle will do every 1 second, `dt` will always be way below 1 second unless there is immense lag going on, so the rotation will accumulate over the lifetime of the particle

src/modifier/rotate_over_time.rs
hukasu: Replaced it for a Vec3 with euler angles, also have the advantage of being able to represent rotations of over 360 degrees

Cargo.toml
hukasu: which example would it be ok to include this modifier then? if any?

PR #554, opened 2026-07-27
Fix two `SimulationSpace::Local` transform bugs

x46085:
TL;DR: In pursuit of these Elodin features ([new example](https://github.com/elodin-sys/elodin/pull/745), [effects tool](https://github.com/elodin-sys/pyrotechnique)), I discovered 2 transform bugs that manifest with emitters that are very far from the origin. This PR attempts to fix them in a maintainer acceptable manner. Happy to adjust as directed.

## Summary

Two independent bugs in how `SimulationSpace::Local` handles the emitter
transform. Both are invisible when the emitter sits at the world origin without
rotation, which is how every example is set up, and both become obvious as soon
as the emitter moves or turns:

1. The camera is converted into effect space with only the rotation part of the
   inverse emitter transform, so camera-oriented particles stop tracking the
   camera as the emitter moves away from the world origin.
2. The init pass applies the emitter transform to shape modifiers and the render
   pass applies it again, so a rotated local-space emitter spawns its shape
   rotated twice.

I found these building a spacecraft visualization where the emitters are
far from the origin and pitch during flight. Each fix is one commit, with
tests and changelog in their own commits.

## 1. Camera position in effect space ignores the emitter translation

`get_camera_position_effect_space()` reconstructs the inverse emitter transform
from the 3x3 block only:

```wgsl
let inverse_transform = transpose(mat3x3(
    spawners[spawner_index].inverse_transform[0].xyz,
    spawners[spawner_index].inverse_transform[1].xyz,
    spawners[spawner_index].inverse_transform[2].xyz,
));
return inverse_transform * view_pos;
```

An affine inverse is `R^-1 * (cam - T)`; this computes `R^-1 * cam`. In effect
space the camera therefore appears where the *world* camera is, not where it is
relative to the emitter. `OrientMode::FaceCameraPosition` and
`OrientMode::AlongVelocity` end up orienting against `normalize(T + offset)`
instead of `normalize(offset)`, so once `|T|` exceeds the camera distance the
direction barely changes as the camera moves: billboards face a fixed world
direction and can be seen edge-on.

`GpuSpawnerParams::new()` already uploads the full inverse affine, and the
shader already has `unpack_compressed_transform()` for exactly this, so the fix
just uses it.

Repro on top of `main`, putting the stock `billboard` example 10 km from the
origin in local space (camera and ground move with it, so framing is unchanged):

```diff
--- a/examples/billboard.rs
+++ b/examples/billboard.rs
+const FAR: Vec3 = Vec3::new(0.0, 0.0, 10_000.0);
+
 fn setup(
     ...
     commands.spawn((
-        Transform::from_xyz(3.0, 3.0, 3.0).looking_at(Vec3::ZERO, Vec3::Y),
+        Transform::from_translation(FAR + Vec3::new(3.0, 3.0, 3.0)).looking_at(FAR, Vec3::Y),
         Camera3d::default(),
@@
         EffectAsset::new(32768, SpawnerSettings::rate(64.0.into()), module)
             .with_name("billboard")
+            .with_simulation_space(SimulationSpace::Local)
@@
     // The ground
     commands.spawn((
-        Transform::from_xyz(0.0, -0.5, 0.0)
+        Transform::from_translation(FAR + Vec3::new(0.0, -0.5, 0.0))
             * Transform::from_rotation(Quat::from_rotation_x(-FRAC_PI_2)),
@@
     commands.spawn((
         ParticleEffect::new(effect),
+        Transform::from_translation(FAR),
         EffectMaterial {
@@
 fn rotate_camera(...) {
-    **camera_transform =
-        Transform::from_xyz(c * radius_xz, 3.0, s * radius_xz).looking_at(Vec3::ZERO, Vec3::Y)
+    **camera_transform =
+        Transform::from_translation(FAR + Vec3::new(c * radius_xz, 3.0, s * radius_xz))
+            .looking_at(FAR, Vec3::Y)
 }
```

Same patch, same frame, `main` vs this branch:

| before | after |
|---|---|
| billboard-before | billboard-after |

## 2. `FaceCameraPosition` builds its up axis in world space

Adjacent to the above, and in the same commit. The generated code crossed the
(effect-space) view vector with a world-space up vector:

```rust
axis_x = normalize(cross(view.world_from_view[1].xyz, axis_z));
```

With an unrotated emitter the two frames coincide, so this looks fine; with a
rotated one the billboard basis picks up a roll that grows with the rotation.
`get_camera_rotation_effect_space()` already exists and already handles the
local/global split, so the fix is to use it for the up axis.

Note that `get_camera_rotation_effect_space()` itself is left alone: it returns
a rotation, so the 3x3 block is the correct and complete inverse there.

## 3. Local-space init applies the emitter transform twice

Shape modifiers that need an orientation take the spawner transform as a
parameter and apply it with `w = 0`, contributing rotation and scale:

```wgsl
let p2 = transform * vec4<f32>(p, 0.0);
(*particle).position = p2.xyz;
```

`vfx_init.wgsl` passed the world spawner transform for every simulation space.
For `SimulationSpace::Local` the render pass then applies the same transform
again, so particles are stored at `R*p` and drawn at `R*(R*p)`. With an
unrotated emitter `R` is identity and `R*R == R`, which is why this has gone
unnoticed.

Affected modifiers are the ones that receive the transform:
`SetPositionCone3dModifier`, `SetVelocityCircleModifier`, and
`SetVelocityTangentModifier`. Global space keeps the world transform, since its
shape modifiers place geometry in world orientation and the generated
simulation-space snippet reads `transform[3].xyz` for the translation.

The init shader had no `LOCAL_SPACE_SIMULATION` shader def, so the fix threads
it through `ParticleInitPipelineKeyFlags` to keep local- and global-space
effects on distinct specialized init pipelines.

Repro on top of `main`, giving the stock `init` example a fixed tilt so the
emitter is already rotated when the particles spawn:

```diff
--- a/examples/init.rs
+++ b/examples/init.rs
 fn rotate_effect(time: Res<Time>, mut query: Query<(&RotateSpeed, &mut Transform)>) {
     for (speed, mut transform) in &mut query {
-        let a = (time.elapsed_secs() * 0.1 * speed.0) * PI;
-        *transform = Transform::from_rotation(Quat::from_rotation_y(a));
+        *transform = Transform::from_rotation(Quat::from_rotation_z(PI / 4.0));
     }
 }
```

The emitter is tilted 45 degrees. Before, the cone is drawn at 90 degrees, i.e.
rotated twice; after, it matches the emitter. The circle and sphere shapes do
not take the transform, and are identical in both, which isolates the cause:

| before | after |
|---|---|
| init-before | init-after |

## Behavior change

This changes how existing local-space effects render whenever the emitter is
rotated, scaled, or far from the origin. Content that was authored around the
old behavior (compensating for a doubled rotation, or for billboards facing the
world origin) will look different. Global-space effects and local-space effects
on an unrotated emitter at the origin are unaffected: every change is inside
`#ifdef LOCAL_SPACE_SIMULATION`.

As a check on that, the unmodified `init` example (local space, animated
rotation) was captured on `main` and on this branch: the two differ by fewer
pixels than two runs of the same build differ from each other, because that
example spawns its particles while the emitter is still at rest.

## Tests

Both are CPU-only, so they run in the no-GPU CI job:

- `spawner_inverse_transform_is_full_affine` locks the contract the render
  shader now relies on: unpacking the uploaded inverse as a 4x4 and applying it
  to the emitter position yields the effect-space origin. Reconstructing only
  the 3x3 block silently drops that.
- `init_pipeline_local_space_shader_def` checks `LOCAL_SPACE_SIMULATION` reaches
  the init pipeline specialization, and is absent otherwise.

The existing `OrientModifier` codegen test also now asserts the generated shader
uses `get_camera_rotation_effect_space()`.

I did not add an example: the bugs need a large translation or a rotation to be
visible, both repros above are a few lines on existing examples, and the
`run-examples` job is already long.

## Verification

Ran the CI matrix locally on macOS (Apple M2 Max, Metal), all passing:

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets --all-features -- -Dwarnings`
- `cargo test --lib --no-default-features --features 2d`, `3d`, and `"2d 3d"`
- `WGPU_BACKEND=metal cargo test --no-default-features --features "2d 3d gpu_tests"`
- `cargo check --lib --all-features` on 1.95 (MSRV)
- `cargo check --target wasm32-unknown-unknown --no-default-features --features 2d,3d`

## Notes for review

- The identity matrix in the local-space init branch keeps a single code path
  through the shape modifiers; it is a constant, so the multiply folds away.
- `get_camera_position_effect_space()` calls `unpack_compressed_transform()`,
  which is defined further down the file. That is valid WGSL (module-scope
  declarations are in scope for the whole program) and naga accepts it, so I
  left the helper where it is rather than reordering and adding diff noise. Say
  the word if you would rather have it moved.
- Happy to split this into two PRs if you prefer to take them separately.

_Disclosure: while this work was done with some AI assistance; the analysis, repros, and
verification above are mine and were run locally and against use in Elodin to verify fix robustness._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed a bug where some GPU operations queued with a buffer reference could be reused
   after the buffer was reallocated, leading to use-after-free (stale data) on GPU.
 - Fixed a double-free of a GPU buffer when despawning a particle effect with a child effect. (#510)
+- Fixed camera-oriented particles (`OrientMode::FaceCameraPosition`, `OrientMode::AlongVelocity`)
+  in `SimulationSpace::Local` effects not tracking the camera when the emitter is away from
+  the world origin or rotated.
+- Fixed `SimulationSpace::Local` effects applying the emitter transform twice to
+  `SetPositionCone3dModifier`, `SetVelocityCircleModifier`, and `SetVelocityTangentModifier`.
+  Local-space effects with a rotated, scaled, or off-origin emitter may render differently
+  than before; this is the corrected behavior.
 
 ## [0.19.0] 2026-06-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,9 +45,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed a bug where some GPU operations queued with a buffer reference could be reused
   after the buffer was reallocated, leading to use-after-free (stale data) on GPU.
 - Fixed a double-free of a GPU buffer when despawning a particle effect with a child effect. (#510)
-- Fixed camera-oriented particles (`OrientMode::FaceCameraPosition`, `OrientMode::AlongVelocity`)
-  in `SimulationSpace::Local` effects not tracking the camera when the emitter is away from
-  the world origin or rotated.
+- Fixed the camera position in effect space used by camera-oriented particles
+  (`OrientMode::FaceCameraPosition`, `OrientMode::AlongVelocity`) in
+  `SimulationSpace::Local` effects: the emitter translation was ignored, most noticeable when emitter is far from origin.
 - Fixed `SimulationSpace::Local` effects applying the emitter transform twice to
   `SetPositionCone3dModifier`, `SetVelocityCircleModifier`, and `SetVelocityTangentModifier`.
   Local-space effects with a rotated, scaled, or off-origin emitter may render differently
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -518,10 +518,9 @@ pub enum SimulationSpace {
 impl SimulationSpace {
     /// Evaluate the simulation space expression.
     ///
-    /// - In the init context, [`SimulationSpace::Global`] converts the particle
-    ///   from emitter (spawn) space into simulation space. Local is a no-op.
-    /// - In the update context, this expression transforms the particle's
-    ///   position from simulation space to storage space.
+    /// - Init: [`SimulationSpace::Global`] converts emitter space to simulation
+    ///   space; [`SimulationSpace::Local`] is a no-op.
+    /// - Update: no-op. Particles are already in simulation space.
     /// - In the render context, this expression transforms the particle's
     ///   position from simulation space to view space.
     pub fn eval(&self, context: &dyn EvalContext) -> Result<String, ExprError> {
@@ -547,18 +546,7 @@ impl SimulationSpace {
                 }
                 SimulationSpace::Local => Ok("".to_string()),
             },
-            ModifierContext::Update => match *self {
-                SimulationSpace::Global => {
-                    if !context.particle_layout().contains(Attribute::POSITION) {
-                        return Err(ExprError::GraphEvalError(format!("Global-space simulation requires that the particles have a {} attribute.", Attribute::POSITION.name())));
-                    }
-                    Ok(format!(
-                        "particle.{} += transform[3].xyz;", // TODO: get_view_position()
-                        Attribute::POSITION.name()
-                    ))
-                }
-                SimulationSpace::Local => Ok("".to_string()),
-            },
+            ModifierContext::Update => Ok(String::new()),
             ModifierContext::Render => Ok(match *self {
                 // TODO: cast vec3 -> vec4 auomatically
                 SimulationSpace::Global => "vec4<f32>(local_position, 1.0)",
@@ -2423,9 +2411,9 @@ else { return c1; }
         let property_layout = PropertyLayout::default();
         let texture_layout = TextureLayout::default();
         {
-            // Local is always available
+            // Global init requires storing the particle's position
             let ctx = ShaderWriter::new(
-                ModifierContext::Update,
+                ModifierContext::Init,
                 &property_layout,
                 &particle_layout,
                 &texture_layout,
@@ -2433,10 +2421,9 @@ else { return c1; }
             assert!(SimulationSpace::Local.eval(&ctx).is_ok());
             assert!(SimulationSpace::Global.eval(&ctx).is_err());
 
-            // Global requires storing the particle's position
             let particle_layout = ParticleLayout::new().append(Attribute::POSITION).build();
             let ctx = ShaderWriter::new(
-                ModifierContext::Update,
+                ModifierContext::Init,
                 &property_layout,
                 &particle_layout,
                 &texture_layout,
@@ -2445,17 +2432,15 @@ else { return c1; }
             assert!(SimulationSpace::Global.eval(&ctx).is_ok());
         }
         {
-            // Local is always available
             let ctx = ShaderWriter::new(
                 ModifierContext::Update,
                 &property_layout,
                 &particle_layout,
                 &texture_layout,
             );
-            assert!(SimulationSpace::Local.eval(&ctx).is_ok());
-            assert!(SimulationSpace::Global.eval(&ctx).is_err());
+            assert_eq!(SimulationSpace::Local.eval(&ctx).unwrap(), "");
+            assert_eq!(SimulationSpace::Global.eval(&ctx).unwrap(), "");
 
-            // Global requires storing the particle's position
             let particle_layout = ParticleLayout::new().append(Attribute::POSITION).build();
             let ctx = ShaderWriter::new(
                 ModifierContext::Update,
@@ -2463,8 +2448,8 @@ else { return c1; }
                 &particle_layout,
                 &texture_layout,
             );
-            assert!(SimulationSpace::Local.eval(&ctx).is_ok());
-            assert!(SimulationSpace::Global.eval(&ctx).is_ok());
+            assert_eq!(SimulationSpace::Local.eval(&ctx).unwrap(), "");
+            assert_eq!(SimulationSpace::Global.eval(&ctx).unwrap(), "");
         }
         {
             // In the render context, the particle position is always available (either
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -518,13 +518,36 @@ pub enum SimulationSpace {
 impl SimulationSpace {
     /// Evaluate the simulation space expression.
     ///
-    /// - In the init and udpate contexts, this expression transforms the
-    ///   particle's position from simulation space to storage space.
+    /// - In the init context, [`SimulationSpace::Global`] converts the particle
+    ///   from emitter (spawn) space into simulation space. Local is a no-op.
+    /// - In the update context, this expression transforms the particle's
+    ///   position from simulation space to storage space.
     /// - In the render context, this expression transforms the particle's
     ///   position from simulation space to view space.
     pub fn eval(&self, context: &dyn EvalContext) -> Result<String, ExprError> {
         match context.modifier_context() {
-            ModifierContext::Init | ModifierContext::Update => match *self {
+            ModifierContext::Init => match *self {
+                SimulationSpace::Global => {
+                    if !context.particle_layout().contains(Attribute::POSITION) {
+                        return Err(ExprError::GraphEvalError(format!("Global-space simulation requires that the particles have a {} attribute.", Attribute::POSITION.name())));
+                    }
+                    let pos = Attribute::POSITION.name();
+                    let mut code = format!(
+                        "particle.{0} = (transform * vec4<f32>(particle.{0}, 1.0)).xyz;\n",
+                        pos
+                    );
+                    if context.particle_layout().contains(Attribute::VELOCITY) {
+                        let vel = Attribute::VELOCITY.name();
+                        code.push_str(&format!(
+                            "particle.{0} = (transform * vec4<f32>(particle.{0}, 0.0)).xyz;\n",
+                            vel
+                        ));
+                    }
+                    Ok(code)
+                }
+                SimulationSpace::Local => Ok("".to_string()),
+            },
+            ModifierContext::Update => match *self {
                 SimulationSpace::Global => {
                     if !context.particle_layout().contains(Attribute::POSITION) {
                         return Err(ExprError::GraphEvalError(format!("Global-space simulation requires that the particles have a {} attribute.", Attribute::POSITION.name())));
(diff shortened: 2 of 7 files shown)


src/render/vfx_init.wgsl
```diff
@@ -159,7 +159,21 @@ fn main(@builtin(global_invocation_id) global_invocation_id: vec3<u32>) {
...
     seed = pcg_hash(particle_index ^ (*spawner).seed);
 
-    // Spawner transform
+    // Spawner transform applied by shape modifiers (cone/sphere/...).
+    // Local-space simulation stores particle attributes in effect space and
+    // applies the emitter transform only at render time. Using the world
+    // spawner matrix here would bake R into positions that render then
+    // multiplies by R again (double rotation) — pitched Local emitters lean
+    // off the geometric axis. Global-space init still needs the world matrix
+    // so shapes land in world before SIMULATION_SPACE_TRANSFORM_PARTICLE.
```
djeedai: AI-generated comment describing a bug. Comment should never describe bugs, especially already-fixed ones. This makes absolutely no sense for the reader who's unaware of the bug.

src/render/vfx_init.wgsl
```diff
@@ -159,7 +159,21 @@ fn main(@builtin(global_invocation_id) global_invocation_id: vec3<u32>) {
...
+#ifdef LOCAL_SPACE_SIMULATION
+    let transform = mat4x4<f32>(
```
djeedai: Kind of OK if we assume that `transform` is only ever used by the various "position" modifiers. I'd rather we clarify the semantic here and call that `sim_from_spawner` for example, which makes it more clear I think that this transforms a point in spawner space into a point in simulation space (whatever the simulation space config is, global or local). That way if any other shader attempts to use this, they won't assume it's the raw unmodified emitter transform.

src/render/mod.rs
```diff
@@ -1789,6 +1789,9 @@ bitflags! {
         const ATTRIBUTE_PREV = (1u8 << 1);
         const ATTRIBUTE_NEXT = (1u8 << 2);
         const CONSUME_GPU_SPAWN_EVENTS = (1u8 << 3);
+        /// Key: LOCAL_SPACE_SIMULATION — init shape modifiers must not bake the
+        /// world spawner transform into particle attributes (render applies it).
```
djeedai: That AI-generated comment refers to a very specific use case for that flag, which is much more restrictive that what the flag actually does. Once again AI commenting about a bug and not about what the code does. The flag name is pretty self-explanatory here, there's no need to document.

src/render/mod.rs
```diff
@@ -7749,6 +7764,75 @@ mod tests {
...
+        );
+
+        let world_point = Vec3::new(1.0, 2.0, 3.0);
+        let expected = transform.affine().inverse().transform_point3(world_point);
+        let actual = inverse.transform_point3(world_point);
+        assert!(
+            actual.abs_diff_eq(expected, 1e-3),
+            "expected {expected:?}, got {actual:?}"
+        );
+    }
```
djeedai: All of this is "testing" some test-only CPU code that is never executed on the GPU, or even anywhere else in Hanabi. This is the classic "the AI agent has no idea how to write a GPU test so makes a similar CPU test", which is completely useless to prevent GPU bugs since the two codes are unrelated and can't be kept in sync (they're not even in the same language), and adds some useless maintenance cost for a test that tests itself.

src/render/mod.rs
```diff
@@ -7749,6 +7764,75 @@ mod tests {
...
+
+    /// Local- and global-space effects must not share a specialized init
+    /// pipeline: the shader def decides whether shape modifiers receive the
+    /// emitter transform or an identity matrix.
```
djeedai: Not very descriptive; I suggest removing.

CHANGELOG.md
```diff
@@ -17,6 +17,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  direction that drifted toward the world origin as the emitter moved away from it, so they
+  stopped tracking the camera and could render edge-on. `OrientMode::FaceCameraPosition` also
+  built its up axis from the world-space view matrix instead of the effect-space camera
+  rotation, adding a roll on rotated emitters.
+- Fixed `SimulationSpace::Local` effects applying the emitter transform twice to shape
+  modifiers: once in the init pass and again when rendering. `SetPositionCone3dModifier`,
+  `SetVelocityCircleModifier`, and `SetVelocityTangentModifier` are affected; a rotated
+  emitter sheared its spawn shape off axis and a scaled one scaled it twice. Note that local
+  space effects on emitters which are rotated, scaled, or far from the origin will render
+  differently than before; this is the intended, corrected behavior.
```
djeedai: A bit too verbose for a CHANGELOG. Just explain briefly what was fixed and how it may change the behavior. The details of the bugs are irrelevant for users reading a CHANGELOG, they only want to know if some bug that might have affected them is fixed, and/or if there's any behavior change they should pay attention to. This is not a PR review, we're not talking to Hanabi maintainers here.

djeedai (changes requested):
Thanks for the change. OK in principle, looks like bugs indeed, but the AI generated docs/comments is a bit too much. Please do a pass manually and review. AI is particularly bad at comments. Also I'd like to improve the semantic of the transform variable name so we avoid future bugs. See review comments for details.

src/render/vfx_init.wgsl
x46085: removed

src/render/vfx_init.wgsl
x46085: updated to "sim_from_spawner", re-ran modifier codegen, re-tested against the same two examples above as well as the downstream Elodin features using it

src/render/mod.rs
x46085: removed

src/render/mod.rs
x46085: removed

src/render/mod.rs
x46085: removed

CHANGELOG.md
x46085: tightened this up

x46085:
Greatly appreciate you taking the time to review, totally agree with the feedback. I went through and removed all extraneous comments, updated the semantics as you suggested, ran codegen and re-tested everywhere as before. Thanks for all you the work you put into this project! We are big fans.

djeedai:
Sorry I've not forgotten about this, change looks good, I just want to confirm the math and I haven't found some time to do so yet.

src/render/vfx_init.wgsl
```diff
@@ -159,15 +159,24 @@ fn main(@builtin(global_invocation_id) global_invocation_id: vec3<u32>) {
...
+#else
+    let sim_from_spawner = transpose(
         mat4x4(
             (*spawner).transform[0],
             (*spawner).transform[1],
             (*spawner).transform[2],
             vec4<f32>(0.0, 0.0, 0.0, 1.0)
         )
     );
+#endif
```
djeedai: The more I think about it the more this looks wrong.

- In local-space simulation, particles are simulated in emitter space. So the particle position inside the GPU buffer is in local space. Only at render time do you need to convert that to global space.
- In global-space simulation, particles are simulated in global space. The particle position inside the GPU buffer is already in global space; the emitter transform is irrelevant after spawning.

So there's no case where the init pass needs to know about the spawner transform.

The only case where this is relevant is for some specific modifiers or expressions. For example, if you pass a world-space position into a property, but do a local-space simulation, then you'd need to convert global -> local inside the pass. But that's probably a bad example anyway because you'd better transform on the CPU side once into local space, rather than once per particle on the GPU.

So I really don't see where you need any transform in Init or Update.

And the bug of applying twice the spawner transform then makes sense : it's applied in the Render pass (as expected) and in the Init pass (bug). Just deleting that from the few modifiers emitting the wrong code should be enough, without touching the shader itself.

x46085: ah, yes that's cleaner — init shouldn’t special-case simulation space. I can drop the identity-matrix ifdef 

One issue I want to point out before I fix: global CPU spawn still needs a conversion. That’s already position += transform[3].xyz in SimulationSpace::eval; the modifier multiply was the rotation/scale half of the same thing. Deleting only that would leave global cones world-aligned (spawn.rs rotates one on purpose). 

If acceptable, I can fold it into that existing snippet so modifiers stay in emitter space, and local eval will stay empty. Lmk if that works for you

djeedai: Yes I think that makes sense:
- particles are spawned in the emitter space
- they're stored and simulated in whatever space the effect configures
- if those are different (so, if they simulate in global space), then the Init pass needs to apply a transform spawn -> sim on init

Is that what you meant?

x46085: Yea, modifiers write emitter space, and SimulationSpace::eval applies spawn -> sim only for global. Local is a no-op there. Thanks for the quick reply, I'll get on it shortly

src/lib.rs
```diff
@@ -518,13 +518,36 @@ pub enum SimulationSpace {
 impl SimulationSpace {
     /// Evaluate the simulation space expression.
     ///
-    /// - In the init and udpate contexts, this expression transforms the
-    ///   particle's position from simulation space to storage space.
+    /// - In the init context, [`SimulationSpace::Global`] converts the particle
+    ///   from emitter (spawn) space into simulation space. Local is a no-op.
+    /// - In the update context, this expression transforms the particle's
+    ///   position from simulation space to storage space.
```
djeedai: The simulation and storage space are by definition the same one. There should be nothing to do in the Update pass.

CHANGELOG.md
```diff
@@ -45,6 +45,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed a bug where some GPU operations queued with a buffer reference could be reused
   after the buffer was reallocated, leading to use-after-free (stale data) on GPU.
 - Fixed a double-free of a GPU buffer when despawning a particle effect with a child effect. (#510)
+- Fixed camera-oriented particles (`OrientMode::FaceCameraPosition`, `OrientMode::AlongVelocity`)
+  in `SimulationSpace::Local` effects not tracking the camera when the emitter is away from
+  the world origin or rotated.
```
djeedai: That's a bit incorrect; it never tracks the camera position, it's just that it's more noticeable when away from the world origin.

djeedai (changes requested):
Still perplex about that `SimulationSpace` having any effect in Update pass; update pass reads from storage, updates, and write back. Everything happens in the simulation space. There's no separate storage space; by definition particles are stored in the simulation space. "Storage space" was just for me to try and explain, but that doesn't exist in the code.

src/lib.rs
x46085: fixed

CHANGELOG.md
x46085: fixed

x46085:
> Still perplex about that `SimulationSpace` having any effect in Update pass; update pass reads from storage, updates, and write back. Everything happens in the simulation space. There's no separate storage space; by definition particles are stored in the simulation space. "Storage space" was just for me to try and explain, but that doesn't exist in the code.

You were right, I was able to remove that aspect, and retested with all of my big_space leveraging examples in [pyrotechnique](https://github.com/elodin-sys/pyrotechnique) and Elodin, and seems to be fine without it. Back over to you.

PR #558, opened 2026-08-09
fix pink screen in Metal HDR examples

jsrgb:
Optional workaround for https://github.com/bevyengine/bevy/issues/25344

Screenshot:
image

Examples that use HDR now start normally


djeedai (commented):
If that's the "fix" then it's very clearly a Bevy bug.

